PR #69, opened 2022-02-12
feat: add cached_async convenience decorator

kvcache:
```
@cached_async(momento_client, cache='cache_name', ttl_seconds=90)
async def load_from_database(key) -> str:
  [...]
```
and you get a view of your database with a 90 second pressure relief
by calling `await load_from_database(key)`

Does not coordinate for thundering herd, just does a basic
hit or miss-load-set.

is_class_method parameter is a bit unfortunate but it's necessary to
disambiguate between `self` as a cache parameter and an object as the
first parameter (which would be used if you were providing a custom
serializer and deserializer)

Do we want to put tools like this in the SDK repo or do we want to make a separate "quality of life" package?

I made a toy with this; it simulates a 2 second database load with a 90 second cache ttl and randomly tries to get an item out of a key space of 1000 and a limited max concurrency of 32. Without the cache, it peaks at 16hz, because it takes 2 seconds to load from the database. With the cache this toy settles around 200hz because many of the requests get serviced from Simple Cache instead of through the 2 second loader.
```
window: 1172 | done:       1681 | rate:  168.0hz | concurrency:  31 | p999:  37.9ms | p90:  19.4ms | p50:  16.4ms | hit:  92.52% | throttle:   0.00% | err:   0.00%         
window: 1173 | done:       1798 | rate:  179.6hz | concurrency:  31 | p999:  50.8ms | p90:  18.7ms | p50:  16.2ms | hit:  92.52% | throttle:   0.00% | err:   0.00%         
window: 1174 | done:       1885 | rate:  188.4hz | concurrency:  31 | p999:  34.1ms | p90:  18.8ms | p50:  16.2ms | hit:  92.52% | throttle:   0.00% | err:   0.00%         
window: 1175 | done:       1960 | rate:  195.9hz | concurrency:  31 | p999:  38.1ms | p90:  18.9ms | p50:  16.5ms | hit:  92.52% | throttle:   0.00% | err:   0.00%         
window: 1176 | done:       2047 | rate:  204.5hz | concurrency:  31 | p999:  38.1ms | p90:  20.6ms | p50:  16.8ms | hit:  92.52% | throttle:   0.00% | err:   0.00%         
window: 1177 | done:       2010 | rate:  200.8hz | concurrency:  28 | p999:  34.6ms | p90:  19.7ms | p50:  16.4ms | hit:  92.52% | throttle:   0.00% | err:   0.00%         
window: 1178 | done:       2066 | rate:  204.7hz | concurrency:  32 | p999:  36.3ms | p90:  20.7ms | p50:  16.5ms | hit:  92.52% | throttle:   0.00% | err:   0.00%         
window: 1179 | done:       2202 | rate:  220.2hz | concurrency:  31 | p999:  35.2ms | p90:  20.2ms | p50:  16.5ms | hit:  92.52% | throttle:   0.00% | err:   0.00% 
```

### DIFF
--- a/src/momento/tools/cached_async.py
+++ b/src/momento/tools/cached_async.py
@@ -1,0 +1,132 @@
+import functools
+
+from ..aio.simple_cache_client import SimpleCacheClient
+from ..cache_operation_responses import CacheGetStatus
+from ..errors import SdkError
+import json
+import time
+import typing
+
+
+_TReturn = typing.TypeVar("_TReturn")
+_TArg = typing.TypeVar("_TArg")
+
+
+def cached_async(
+    client: SimpleCacheClient,
+    cache: str,
+    *,
+    is_class_method: bool = False,
+    ttl_seconds: float = None,
+    key_prefix: str = None,
+    from_string_or_bytes=json.loads,
+    to_string_or_bytes=json.dumps,
+    on_hit: typing.Callable[[int], None] = lambda load_latency_ns: None,
+    on_miss: typing.Callable[
+        [int, int, int], None
+    ] = lambda load_latency_ns, store_latency_ns, compute_latency_ns: None,
+    on_load_error: typing.Callable[[SdkError], None] = lambda error: None,
+    on_store_error: typing.Callable[[SdkError], None] = lambda error: None,
+    fail_open: bool = True,
+):
+    """
+    Decorator for a function that returns an expensive value from Momento Simple Cache.
+
+    If you have a simple function that caches a string, you'll just do:
+
+    @cached_async(client, 'function_cache', ttl_seconds=90)
+    async def load_from_database(user: str, attribute: str) -> str:
+        user_object = await database.read(table='users', user_id=user)
+        return user_object[attribute]
+
+    This will cache each user's accessed attributes for 90 seconds. Hot attributes will
+    tend to be in the cache and attributes displayed to the user won't be very stale.
+
+    :param client: your configured Simple Cache client
+    :param cache: Cache to use for this function's values
+    :param is_class_method: [default=False] OPTIONAL set this True if you're decorating something that uses self
+    :param ttl_seconds: OPTIONAL how long to store answers in Simple Cache
+    :param key_prefix: [default=function_name] OPTIONAL prefix for keys in the cache to separate them from other cached
+                       functions in the same shared cache
+    :param from_string_or_bytes: OPTIONAL strategy for loading directly into an object. If you are storing
+                                 complex objects you'll need to provide a deserializer here. You can use json,
+                                 protocol buffers, whatever you want
+    :param to_string_or_bytes: OPTIONAL strategy for writing cache values. If you are storing complex objects you'll
+                               need to provide a serializer here. You can use json, protocol buffers, whatever you want
+    :param on_hit: OPTIONAL callback for metrics
+    :param on_miss: OPTIONAL callback for metrics
+    :param on_load_error: OPTIONAL callback for metrics and tracing
+    :param on_store_error: OPTIONAL callback for metrics and tracing
+    :param fail_open: [default=True] OPTIONAL set False if you want the function to raise on error instead of invoking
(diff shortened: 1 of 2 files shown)


cprice404:
This is really cool and I love the idea of building up some stuff like this for various languages.

I think that my instinct would be to *not* put this directly into the SDK for now though; I think the SDK probably should be kept to the bare-minimum set of functionality so that it is as easy as humanly possible to monitor/maintain backward compatibility guarantees, etc.  I can imagine stuff like this going in an `-extras` or `-goodies` add-on library for each language, or just going into the `client-sdk-examples` repo for now or something.

I don't feel *too* strongly about that, could be persuaded that it would be good to put stuff like this directly in the SDKs but based on past experience my gut would be to minimize surface area for official SDKs.  Would be happy to discuss!

And I definitely think that if this *doesn't* land in this repo, it'd be fair game to link to it in the README of this repo.

kvcache:
> I don't feel *too* strongly about that, could be persuaded that it would be good to put stuff like this directly in the SDKs but based on past experience my gut would be to minimize surface area for official SDKs.  Would be happy to discuss!

I'm with you - we can find some parking lot time to chat about it with interested parties. I was hoping to spark this conversation!

eaddingtonwhite:
@kvcache @cprice404 @rlinehan @gautamomento coming back to revisit this as it's been a while. I recently did POC of adding Momento as a driver to Dogpile cache in python. 
See https://github.com/eaddingtonwhite/momento-dogpile

I would like to contribute it down as a backend option into [main dogpile repo.](https://github.com/sqlalchemy/dogpile.cache) I just wanted to wait till we got initial customer feedback from the customer doing POC I did this for before. I'm hoping though in the next couple of weeks can make it happen.

Dogpile has a decorator function built into it already so can use like this.
```python
@someregion.cache_multi_on_arguments()
def generate_something(*keys):
    return [
        somedatabase.query(key)
        for key in keys
    ]
```
https://dogpilecache.sqlalchemy.org/en/latest/api.html?highlight=decorator#dogpile.cache.region.CacheRegion.cache_multi_on_arguments 

Thinking this might be a better approach for us to start keeping our SDK repo thinner and less for us to maintain for now. What do you both think? Would we be ok closing this then if we go with that approach for now? We could maybe add some links/examples in README.md for python SDK to show users how to enable this?

cprice404:
> @kvcache @cprice404 @rlinehan @gautamomento coming back to revisit this as it's been a while. I recently did POC of adding Momento as a driver to Dogpile cache in python. See https://github.com/eaddingtonwhite/momento-dogpile
> 
> I would like to contribute it down as a backend option into [main dogpile repo.](https://github.com/sqlalchemy/dogpile.cache) I just wanted to wait till we got initial customer feedback from the customer doing POC I did this for before. I'm hoping though in the next couple of weeks can make it happen.
> 
> Dogpile has some decorator function's built into it already so can use like this.
> 
> ```python
> @someregion.cache_on_arguments()
> def generate_something(x, y):
>     return somedatabase.query(x, y)
> 
> @region.cache_on_arguments(namespace="foo")
> def one(a, b):
>     return a + b
> ```
> 
> https://dogpilecache.sqlalchemy.org/en/latest/api.html?highlight=decorator#dogpile.cache.region.CacheRegion.cache_on_arguments
> 
> ```python
> @someregion.cache_multi_on_arguments()
> def generate_something(*keys):
>     return [
>         somedatabase.query(key)
>         for key in keys
>     ]
> ```
> 
> https://dogpilecache.sqlalchemy.org/en/latest/api.html?highlight=decorator#dogpile.cache.region.CacheRegion.cache_multi_on_arguments
> 
> Thinking this might be a better approach for us to start keeping our SDK repo thinner and less for us to maintain for now. What do you both think? Would we be ok closing this then if we go with that approach for now? We could maybe add some links/examples in README.md for python SDK to show users how to enable this?

I agree that the home for various integrations like these is probably outside of our SDK repos, and we can link them in examples/docs.

eaddingtonwhite:
@cprice404 Ok going to close this for now. Will revisit down line where right home is.

PR #75, opened 2022-03-31
chore: Add `agent` metadata

### LINKED ISSUE #74: Add `agent` metadata to `AuthInterceptor` when starting a call
https://github.com/momentohq/client-sdk-java/issues/124

poppoerika:
Closes #74

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,6 @@ install_requires =
     setuptools
     pyjwt
     grpcio
-    click==8.0.4
 
 [options.packages.find]
 where = src
@@ -27,3 +26,6 @@ mypy =
 
 black =
     black==22.1.0
+
+click =
+    click==8.0.4
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,4 @@ mypy =
     mypy==0.931
 
 black =
-    black==22.1.0
-
-click =
-    click==8.0.4
+    black==22.3.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,6 +16,7 @@ install_requires =
     setuptools
     pyjwt
     grpcio
+    click==8.0.4
 
 [options.packages.find]
 where = src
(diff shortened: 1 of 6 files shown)


eaddingtonwhite (approved):

setup.cfg
```diff
@@ -16,6 +16,7 @@ install_requires =
     setuptools
     pyjwt
     grpcio
+    click==8.0.4
```
eaddingtonwhite, line 19: Is this customer facing pacakge? Do we use this just for testing ? Should it maybe live down in `extras_require` section with `mypy` and `black`?

poppoerika: Removed that and updated the black module to the latest version based on this post: https://github.com/psf/black/issues/2964

eaddingtonwhite (dismissed):

src/momento/aio/_add_header_client_interceptor.py
```diff
@@ -20,6 +34,11 @@ async def intercept_unary_unary(
     ) -> Union[grpc.aio._call.UnaryUnaryCall, grpc.aio._typing.ResponseType]:
         if client_call_details.metadata is None:
             client_call_details.metadata = Metadata()
-        client_call_details.metadata.add(self._header_name, self._header_value)
+        for header in self.headers_to_add_every_time:
+            client_call_details.metadata.add(header.name, header.value)
+        if AddHeaderClientInterceptor.are_only_once_headers_sent == False:
+            for header in self._headers_to_add_once:
```
eaddingtonwhite, line 40: this seems much safer and cleaner then the `[0]` reference before 👍

eaddingtonwhite (approved):

src/momento/aio/_add_header_client_interceptor.py
poppoerika: Yup, as @gautamomento in the csharp sdk pr, I'm trying to implement future proof option rather than fixed index here.

PR #78, opened 2022-04-07
feat: add support to create pre-signed URLs

virratanasangpunth:


src/momento/momento_signer.py
```diff
@@ -0,0 +1,140 @@
...
+        Args:
+            jwk_json_string: the JSON string of the JWK key. This can be obtained from create_signing_key response.
+
+        Raises:
+            InvalidArgumentError: If the supplied private key is not valid.
+        """
+        try:
+            jwk_json: Dict[str, str] = json.loads(jwk_json_string)
+        except json.decoder.JSONDecodeError:
+            raise InvalidArgumentError(f"Invalid JWK Json String: {jwk_json_string}")
```
kvcache, line 68: All of the exceptions in this PR swallow root causes which could be helpful. Here is an example of how to raise exceptions with additional context in Python:
```
>>> def f():
...  raise Exception('it failed')
... 
>>> try:
...  f()
... except Exception as e:
...  raise Exception('pay respects') from e
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 2, in f
Exception: it failed

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
Exception: pay respects
```

src/momento/momento_signer.py
```diff
@@ -0,0 +1,140 @@
...
+
+        if "kid" not in jwk_json:
+            raise InvalidArgumentError(f"JWK missing kid attribute: {jwk_json_string}")
+
+        if "alg" not in jwk_json:
+            raise InvalidArgumentError(f"JWK missing alg attribute: {jwk_json_string}")
```
kvcache, line 74: This is not illegal. It could have crv and/or kty.

src/momento/momento_signer.py
```diff
@@ -0,0 +1,140 @@
...
+
+        try:
+            self._jwk: PyJWK = PyJWK.from_dict(jwk_json)  # type: ignore[no-untyped-call, misc]
+        except jwt.exceptions.PyJWKError:
+            raise InvalidArgumentError(f"Invalid JWK: {jwk_json_string}")
+
+        self._kid = jwk_json["kid"]
+        self._alg = jwk_json["alg"]
```
kvcache, line 82: These properties are parsed or inferred by PyJwk already and it is the source of truth for them. We should not store these separately.

src/momento/momento_signer.py
```diff
@@ -0,0 +1,140 @@
...
+class SigningRequest:
+    def __init__(
+        self,
+        cache_name: str,
+        cache_key: str,
+        cache_operation: CacheOperation,
+        expiry_epoch_seconds: int,
+        ttl_seconds: Optional[int] = None,
+    ):
+        """Initializes a new SigningRequest.
```
kvcache, line 24: nice docs

src/momento/momento_signer.py
```diff
@@ -0,0 +1,140 @@
...
+        self._alg = jwk_json["alg"]
+
+    def sign_access_token(self, signing_request: SigningRequest) -> str:
+        """Creates an access token to be used as a JWT token.
+
+        Args:
+            signing_request: Contains parameters for the generated token.
+
+        Returns:
+            str
```
kvcache, line 91: The return type is already in the function signature's return hint. This `Returns:` should say something meaningful or should be removed.

src/momento/momento_signer.py
```diff
@@ -0,0 +1,140 @@
...
+        self, hostname: str, signing_request: SigningRequest
+    ) -> str:
+        """Creates a pre-signed HTTPS URL.
+
+        Args:
+            hostname: Hostname of SimpleCacheService. This value can be obtained from create_signing_key response.
+            signing_request: Contains parameters for the generated URL.
+
+        Returns:
+            str
```
kvcache, line 126: same "use it or lose it" comment here

src/momento/momento_signer.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        token = self.sign_access_token(signing_request)
+        if signing_request.cache_operation() == CacheOperation.GET:
+            return f"https://{hostname}/cache/get/{signing_request.cache_name()}/{signing_request.cache_key()}?token={token}"
```
kvcache, line 130: Are you sure you don't want to url encode the cache key?
```
➜  cacheservice git:(envoydoc) ✗ momento cache set --name '.-_.-_.-' --key '///' --value 'you know' --ttl 90
[2022-04-07T23:29:05Z INFO  momento::commands::cache::cache_cli] setting key: /// into cache: .-_.-_.-
[2022-04-07T23:29:05Z INFO  momento::commands::cache::cache_cli] set success
```

This clearly makes no sense at all:
```
➜  cacheservice git:(envoydoc) ✗ curl https://cache.cell-alpha-dev.preprod.a.momentohq.com/cache/get/.-_.-_.-////\?token\=$JWS_ALPHA 
```

But this works:
```
curl https://cache.cell-alpha-dev.preprod.a.momentohq.com/cache/get/.-_.-_.-/%2f%2f%2f\?token\=$JWS_ALPHA
you know2
```

kvcache (commented):
Looking good. Got a couple things to consider

src/momento/momento_signer.py
virratanasangpunth: Python is exception-chaining for you automatically: https://docs.python.org/3.10/tutorial/errors.html#exception-chaining

oooooo wait a minute.... Looks like this is only for Python3.9+

Will update.

src/momento/momento_signer.py
virratanasangpunth: Yess.... good catch. 

I think we have this bug in c# sdk as well. Ticket filed: https://github.com/momentohq/client-sdk-csharp/issues/43

src/momento/momento_signer.py
virratanasangpunth: `kid` removed.

See the discussion above for `alg`.

src/momento/momento_signer.py
virratanasangpunth: Hmm this is a tricky one. We know the JWKs we return from our service do have 'alg' defined though.

pyjwt library that we are using does have the logic to handle when `alg` is missing but it's not exposed to us unfortunately: https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwk.py#L19

I might need to copy-n-paste that logic into our sdk here.

wandaitzuchen (approved):
lgtm!

mjindalmotiv8 (approved):

kvcache (approved):

PR #81, opened 2022-04-20
feat: add support for CreateSigningKey, RevokeSigningKey, and ListSigningKeys APIs

tylerburdsall:
Adds support for new APIs

### DIFF
--- a/src/momento/aio/_scs_control_client.py
+++ b/src/momento/aio/_scs_control_client.py
@@ -1,13 +1,23 @@
+from time import time
 from typing import Optional
 
 from momento_wire_types.controlclient_pb2 import _CreateCacheRequest
 from momento_wire_types.controlclient_pb2 import _DeleteCacheRequest
 from momento_wire_types.controlclient_pb2 import _ListCachesRequest
+from momento_wire_types.controlclient_pb2 import _CreateSigningKeyRequest
+from momento_wire_types.controlclient_pb2 import _RevokeSigningKeyRequest
+from momento_wire_types.controlclient_pb2 import _ListSigningKeysRequest
 
 from .._utilities._data_validation import _validate_cache_name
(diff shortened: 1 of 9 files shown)


wandaitzuchen (dismissed):
🚢

src/momento/aio/_scs_control_client.py
```diff
@@ -70,5 +80,58 @@ async def list_caches(self, next_token: Optional[str] = None) -> ListCachesRespo
         except Exception as e:
             raise _cache_service_errors_converter.convert(e)
 
+    async def create_signing_key(
+        self, ttl_minutes: int, endpoint: str
```
virratanasangpunth, line 84: I find it a little awkward to pass the `endpoint` as an arg here just to be re-packaged into `CreateSigningKeyResponse` and returned.

If you assign `self._endpoint = endpoint` in the ctor you can use that.

tylerburdsall: This is the data plane endpoint, not the control plane endpoint. That's why I'm passing it differently here

virratanasangpunth: oh right!

virratanasangpunth (approved):

src/momento/aio/_scs_control_client.py
```diff
@@ -70,5 +80,58 @@ async def list_caches(self, next_token: Optional[str] = None) -> ListCachesRespo
...
+    ) -> CreateSigningKeyResponse:
+        _validate_ttl_minutes(ttl_minutes)
+        try:
```
kvcache, line 87: nit: You could make this code read a little nicer with a decorator so you're not copypasting try:except->converter.convert all over the place.

kvcache (approved):

PR #95, opened 2022-05-10
Specify tox environments in format from official documentation.

### LINKED ISSUE #87: Tox virtualenv creation doesn't work in Linux environments
Following the instructions in CONTRIBUTING.md to set up tox virtual environments does not work as expected in Linux. After installing the various python versions, when running `tox --notest`, tox:
- creates virtual environments for python3.7, python3.8, python3.9, and python3.10, but
- each of these installations only contain the python binaries for the system version

malandis:
Closes #87 

The official tox
documentation (https://tox.wiki/en/latest/config.html#tox-environments)
shows python versions in a short format (py37, py38, py39, py1310).

The previous commit to this block (d2342f3) changed it from the
preferred format to python3.x. This does not work in Linux
environments. With version names in the (non-preferred) format, when
running `tox`, the program cannot find the python versions and
defaults to the system python for *every* version. When using the
preferred (short) names, tox works as expected.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,9 @@
 [tox]
 envlist =
-    python3.10
-    python3.9
-    python3.8
-    python3.7
+    py310
+    py39
+    py38
+    py37
     lint
 
 [testenv]


cprice404 (approved):
🚢 as long as someone on a mac has validated!

eaddingtonwhite (approved):

malandis:
> @malandis Tested on my mac after whiping out .tox dir was able to loop through all the versions properly w/ new tox.ini format.

Thanks Ellery. Tyler was also able to reproduce this, so this is good.

PR #97, opened 2022-05-11
chore: Update CONTRIBUTING documentation to reflect latest tox envlist.

malandis:
The previous PR #95 updated the tox envlist to be in the standard format. This PR updates the documentation to be consistent with that.


tylerburdsall (approved):
:shipit:

Issue #92, opened 2022-05-09
Come up with better release pattern for sdk


eaddingtonwhite:
We should have the ability to:
* To bundle up several changes together on a branch before releasing a version.
* Have an easy way to give customer a private version of sdk for feed back with out a public release

malandis:
Are we punting on this for now?

eaddingtonwhite:
I think we can probably de-prioritize since we were going to hold off merging pr till we got api integration done. Would be nice thing to have though eventually. Maybe work on getting interfaces knocked out in c# first for data structures and we can come back to this story last this sprint if have time.

cprice404:
Putting some docs in `standards-and-practices` repo; we can add language-specific info about support for building distributable packages that aren't published to the artifact repos.

cprice404:
https://github.com/momentohq/standards-and-practices/pull/7

PR #98, opened 2022-05-11
feat(#85): MVP dictionary API

### LINKED ISSUE #85: Adds MVP hashmap support to client python sdk

malandis:
Closes #85 

Adds a dictionary data type API with a non-production client side implementation. The methods added to the client are:

1. `dictionary_get`: retrieve a value from a dictionary in the cache
2. `dictionary_get_all`: retrieve the whole dictionary from the cache
3.  `dictionary_set`: store key-value pairs in a dictionary (upsert)

### DIFF
--- a/src/momento/aio/simple_cache_client.py
+++ b/src/momento/aio/simple_cache_client.py
@@ -315,34 +315,33 @@ class SimpleCacheClientIncubating(SimpleCacheClient):
         self,
         cache_name: str,
         dictionary_name: str,
-        mapping: Dictionary,
+        dictionary: Dictionary,
     ) -> CacheDictionarySetResponse:
         """Store dictionary items (key-value pairs) in the cache.
 
-        Inserts items from `mapping` into a dictionary `dictionary_name`.
+        Inserts items from `dictionary` into a dictionary `dictionary_name`.
         Updates (overwrites) values if the key already exists.
 
         Args:
-            cache_name (str): Name of the cache to store the dictionary mapping in.
-            dictionary_name (str): The name of the dictionary to store the mapping.
-            mapping (Dictionary): The items (key-value pairs) to be stored.
+            cache_name (str): Name of the cache to store the dictionary in.
+            dictionary_name (str): The name of the dictionary in the cache.
+            dictionary (Dictionary): The items (key-value pairs) to be stored.
 
         Returns:
             CacheDictionarySetResponse: data stored in the cache
         """
         dictionary_get_response = await self.get(cache_name, dictionary_name)
-        dictionary = {}
+        cached_dictionary = {}
         if dictionary_get_response.status() == CacheGetStatus.HIT:
-            dictionary = cast(
+            cached_dictionary = cast(
                 Dictionary,
                 pickle.loads(cast(bytes, dictionary_get_response.value_as_bytes())),
             )
 
-        mapping = convert_dict_values_to_bytes(mapping)
-        dictionary.update(mapping)
+        cached_dictionary.update(convert_dict_values_to_bytes(dictionary))
 
         set_response = await self.set(
-            cache_name, dictionary_name, pickle.dumps(dictionary)
+            cache_name, dictionary_name, pickle.dumps(cached_dictionary)
         )
         return CacheDictionarySetResponse(
             key=set_response._key, value=_deserialize_stored_hash(set_response._value)
--- a/src/momento/aio/simple_cache_client.py
+++ b/src/momento/aio/simple_cache_client.py
@@ -1,7 +1,5 @@
-import pickle
 from types import TracebackType
-from typing import cast, Optional, Union, Type, List
-import warnings
+from typing import Optional, Union, Type, List
 
 try:
     from ._scs_control_client import _ScsControlClient
@@ -37,37 +35,13 @@ from ..cache_operation_types import (
     ListSigningKeysResponse,
     CacheSetResponse,
     CacheGetResponse,
-    CacheGetStatus,
     CacheMultiSetOperation,
     CacheMultiGetOperation,
     CacheMultiSetFailureResponse,
     CacheMultiSetResponse,
     CacheMultiGetResponse,
     CacheMultiGetFailureResponse,
-    CacheDictionaryGetResponse,
-    CacheDictionarySetResponse,
-    CacheDictionaryValue,
-    CacheDictionaryGetAllResponse,
-    DictionaryKey,
-    Dictionary,
-    StoredDictionary,
 )
-
-
-INCUBATING_WARNING_MSG = "Using the incubating client: functionality and features are experimental and subject to change or deletion!"
-
-
-def convert_dict_values_to_bytes(dict_: Dictionary) -> Dictionary:
-    return {k: v if isinstance(v, bytes) else v.encode() for k, v in dict_.items()}
-
-
-def dict_to_stored_hash(dict_: Dictionary) -> StoredDictionary:
-    return {k: CacheDictionaryValue(v) for k, v in dict_.items()}
-
-
-def _deserialize_stored_hash(pickled_dict: bytes) -> StoredDictionary:
-    d = cast(Dictionary, pickle.loads(pickled_dict))
-    return dict_to_stored_hash(d)
 
 
 class SimpleCacheClient:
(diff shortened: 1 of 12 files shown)


kvcache:
> Closes https://github.com/momentohq/client-sdk-python/issues/85

and

> a non-production client side implementation

does "non-production client side implementation" satisfy "V" in "MVP?"


Also, this is our production sdk. I think you want to put this stuff behind an `incubation()` wrapper or something. There's no guarantee or even a reasonable likelihood that these apis will behave consistently between their `incubation` introduction and their actual implementation.

malandis:
> does "non-production client side implementation" satisfy "V" in "MVP?"
Per the discussion I had regarding issue #85, yes. @eaddingtonwhite @kvcache if this is a problem I can change this to only push stubs (that raise `NotImplementedError`). Thoughts?

kvcache:
I don't see a record of that discussion on the issue. Feel free to link it if you have it so I can be involved.

This is a non-production-ready implementation in our production sdk's production object model namespace. I don't think it should go in quite like this.

eaddingtonwhite:
@malandis I think maybe we swarm on this ticket and come up with better way to have a beta or non public release. 
https://github.com/momentohq/client-sdk-python/issues/92

I do feel we need a way to iterate and try things with customers faster. Alternatively I like @kvcache idea of an `incubation` or `experimental` wrapper or something along those lines so we can have separately as well to start. Maybe could brainstorm off line what that might look like to end user.

malandis:
> @malandis I think maybe we swarm on this ticket and come up with better way to have a beta or non public release. #92

I was thinking back to our zoom parking lot with Daniela earlier this week where we discussed this. I thought it was agreed that because the SDK is experimental and so early on that, in the interest of time (at least for this time), we would push to the main branch. Happy to reconsider -- just want to make sure we're on the same page and meet the timelines.

kvcache:
> because the SDK is experimental and so early on

I do not believe the relevant stakeholders are all aligned on this point.

> Happy to reconsider -- just want to make sure we're on the same page and meet the timelines.

Have you considered, as one possible resolution, how long would it take a developer to add a `SimpleCacheClient##incubation()` function which returned a `SimpleCacheClientIncubation` with WIP & placeholder functionality that is explicitly expected to change?

eaddingtonwhite:
Just chatted w/ @danielamiao a bit on this. I think a bit of disconnect about MVP and what we want to send customers and when. Apologies @malandis I think was just bit telephone between Khawaja, Daniella and I on what we were good sending customers and timing. I will keep working to improve on this for future new higher lvl methods we add to SDK.

 @kvcache I really like your idea of just having an `experimental()` constructor that returns a client w/ some more fwd looking apis on it. I think that is pretty good compromise and also will give us path forward for other things like this in future. 

@malandis after chatting w/ @danielamiao bit more she is good w/ us taking some time today to do this quick before sending to Luiza for feed back. I can make follow up story to make this experimental wrapper client quick. Im fine w/ you adding onto this PR as well if that is easier but open to how ever you think best way to do it. Can chat more offline also.

malandis:
@kvcache @eaddingtonwhite I separated out the new APIs into a separate class, added an incubating flag to the existing `init` methods to toggle to use it (false by default), and a warning (emitted once per process). Let me know your thoughts on this design.

src/momento/simple_cache_client.py
```diff
@@ -284,10 +374,37 @@ def init(
         request_timeout_ms: An optional timeout in milliseconds to allow for Get and Set operations to complete.
             Defaults to 5 seconds. The request will be terminated if it takes longer than this value and will result
             in TimeoutError.
+        incubating (bool): Use the incubating client. Includes non-final, experimental features and APIs.
```
eaddingtonwhite, line 377: I think this can get removed now.

malandis: Good catch!

eaddingtonwhite (approved):
@malandis Nice I think this is looking good to me! I like where we landed w/ incubating and dual factory methods. I think will be nice for devs to explore with. One thought would be maybe we add section to README for how to use SDK functionality that is incubating. Id be ok w/ this as follow up PR though if you want. Maybe have one other person with fresh eyes take one more pass but I think this is good to 🚢 to customers for initial feed back.

src/momento/aio/simple_cache_client.py
```diff
@@ -273,6 +299,105 @@ async def get(self, cache_name: str, key: str) -> CacheGetResponse:
...
+            )
+
+        cached_dictionary.update(convert_dict_values_to_bytes(dictionary))
+
+        set_response = await self.set(
+            cache_name, dictionary_name, pickle.dumps(cached_dictionary)
+        )
+        return CacheDictionarySetResponse(
+            key=set_response._key, value=_deserialize_stored_hash(set_response._value)
+        )
```
kvcache, line 348: nit: same argument complaint, but both different and intensified by the presence of `kwargs`

src/momento/cache_operation_types.py
```diff
@@ -163,6 +163,124 @@ def values_as_bytes(self) -> List[Optional[bytes]]:
         return r_values
 
 
+# Dictionary related respnses
```
kvcache, line 166: s/respnses/responses

src/momento/cache_operation_types.py
```diff
@@ -163,6 +163,124 @@ def values_as_bytes(self) -> List[Optional[bytes]]:
...
+
+    def __str__(self) -> str:
+        return self.__repr__()
+
+    def __repr__(self) -> str:
+        value = self._value
+        try:
+            value = self.value()
+        except UnicodeDecodeError:
+            pass
```
kvcache, line 224: nit: in python you can just
```
try:
  value = self.value()
except UnicodeDecodeError:
  value = self._value
return f"... {value}"
```

here and elsewhere

src/momento/cache_operation_types.py
```diff
@@ -163,6 +163,124 @@ def values_as_bytes(self) -> List[Optional[bytes]]:
...
+
+        return f"CacheDictionaryValue(value={value})"  # type: ignore
+
+
+# Represents the type of a hash as stored in the cache.
+# This is the type returned by hgetall and hset.
```
kvcache, line 230: I think this should be "dictionary_get_all" and "used by dictionary_set" right? Oh, you're modeling sets to return the dictionary too? 🤔

kvcache (commented):
I just spot comments and code style.

src/momento/cache_operation_types.py
malandis: Good catch

kvcache (approved):
Remaining issues:
* Code style is non-compliant & organization is not quite right.
* I'm not sure about the model of returning the dictionary from the dictionary_set() api.

I don't want to keep looping over this on this pr. Looking forward to the cleanup and the maturation of the dictionary api family.

src/momento/cache_operation_types.py
```diff
@@ -163,6 +163,124 @@ def values_as_bytes(self) -> List[Optional[bytes]]:
...
+    def status(self) -> CacheGetStatus:
+        return self._result
+
+    def __str__(self) -> str:
+        return self.__repr__()
+
+    def __repr__(self) -> str:
+        return (
+            f"CacheDictionaryGetAllResponse(value={self._value}, result={self._result})"
+        )
```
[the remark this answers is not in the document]
malandis: Agreed it's odd -- this was done by the black code formatter. The line with the return statement + string is too long (per the settings in black). It's wrapping the string in parens to make it an expression (as opposed to a singleton tuple "(x,)") so it can split it into separate lines. I made some housekeeping issues to revisit this.

src/momento/aio/simple_cache_client.py
```diff
@@ -273,6 +299,105 @@ async def get(self, cache_name: str, key: str) -> CacheGetResponse:
...
+    def __init__(
+        self,
+        auth_token: str,
+        default_ttl_seconds: int,
+        data_client_operation_timeout_ms: Optional[int],
+    ):
+        warnings.warn(INCUBATING_WARNING_MSG)
+        super().__init__(
+            auth_token, default_ttl_seconds, data_client_operation_timeout_ms
+        )
```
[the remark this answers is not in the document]
malandis: Black code formatter again (same as above)

src/momento/aio/simple_cache_client.py
malandis: Black code formatter again

src/momento/cache_operation_types.py
malandis: That's how I had it originally. The mypy (type checking) settings are overly strict and I had to write it like this to pass. Another thing to revisit.

src/momento/cache_operation_types.py
malandis: > Remaining issues:
> 
> * Code style is non-compliant & organization is not quite right.
Code style is consistent with the repo black and mypy settings. I have some housekeeping chores to revisit this to make it less strict. Agreed about the organization and have revised it.

> * I'm not sure about the model of returning the dictionary from the dictionary_set() api.
We can sync up about this

malandis:
> Remaining issues:
> 
> * Code style is non-compliant & organization is not quite right.

Code style is consistent with the repo black and mypy settings. I have some housekeeping chores to revisit this to make it less strict. Agreed about the organization and have revised it.

> * I'm not sure about the model of returning the dictionary from the dictionary_set() api.

We can sync up about this

src/momento/aio/simple_cache_client.py
kvcache: that is one petard by which i tire of being hoist

eaddingtonwhite (approved):
🚢 nice work @malandis looking good!

PR #100, opened 2022-05-18
feat: Add delete functionality

### LINKED ISSUE #84: [New API] Add Delete API
Our backend server now supports delete, this functionality needs to be added into the SDK.

Need to update to use newest proto: https://github.com/momentohq/client_protos/blob/main/proto/cacheclient.proto#L35

malandis:
Closes #84 

Expose delete method to the client. Update momento wire types, hook up gRPC to data client backend, and updates clients.

### DIFF
--- a/src/momento/aio/_scs_data_client.py
+++ b/src/momento/aio/_scs_data_client.py
@@ -1,8 +1,7 @@
 import asyncio
 import functools
 from typing import Union, Optional, List, Any
-from momento_wire_types.cacheclient_pb2 import _GetRequest
-from momento_wire_types.cacheclient_pb2 import _SetRequest
+from momento_wire_types.cacheclient_pb2 import _GetRequest, _SetRequest, _DeleteRequest
 
 from .. import cache_operation_types as cache_sdk_ops
 from .. import _cache_service_errors_converter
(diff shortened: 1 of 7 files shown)


src/momento/aio/_scs_data_client.py
```diff
@@ -234,5 +233,24 @@ async def multi_get(
             _momento_logger.debug(f"multi_get failed with response: {e}")
             raise _cache_service_errors_converter.convert(e)
 
+    async def delete(
+        self, cache_name: str, key: Union[str, bytes]
+    ) -> cache_sdk_ops.CacheDeleteResponse:
+        _validate_cache_name(cache_name)
+        try:
+            _momento_logger.debug(f"Issuing a delete request with key {str(key)}")
```
eaddingtonwhite, line 241: just as im looking at this noticed we could probably DRY up our access logging. Maybe for follow up pr but we could probably extract this debug logging out into some standard decorator vs writing out in every method.

src/momento/simple_cache_client.py
```diff
@@ -269,6 +270,28 @@ def multi_get(
...
+
+        Returns:
+            CacheDeleteResponse
+
+        Raises:
+            InvalidArgumentError: If validation fails for provided method arguments.
+            BadRequestError: If the provided inputs are rejected by server because they are invalid
+            NotFoundError: If the cache with the given name doesn't exist.
+            AuthenticationError: If the provided Momento Auth Token is invalid.
+            InternalServerError: If server encountered an unknown error while trying to retrieve the item.
```
eaddingtonwhite, line 290:
```suggestion
            InternalServerError: If server encountered an unknown error while trying to delete the item.
```

eaddingtonwhite (approved):
Few nits but looks straight forward @malandis nice work! 🚢

src/momento/aio/_scs_data_client.py
kvcache: this is true.


Also, this strategy for logging is not great. You're (1) encoding, (2) copying the key and (3) allocating yet a 3rd copy of the memory just to throw it away probably 99.99999% of the time or more. Why does this need to do all of that work when the `debug` level is not enabled?

kvcache: oh geeze, here & everywhere i see.

eaddingtonwhite: Made issue to revisit. Agree @kvcache

kvcache: you can consider declining to propagate this bad pattern by merging [this fix](https://github.com/momentohq/client-sdk-python/pull/102) and updating your net new logging statements

kvcache: @malandis Did you miss this thread?

src/momento/simple_cache_client.py
kvcache: @malandis did you miss this as well?

PR #113, opened 2022-06-17
chore: Add repr and str methods to response objects

malandis:
Closes #113 

Most of the response objects were lacking `__repr__` methods. Before users could expect output like this:

```python
>>> client.get("my-cache", "my-key")
<momento.cache_client_operations.CacheGetResponse at 0x7f8e6d292770>
```

which is not very helpful. Now we have output like this:

```python
>>> client.get("my-cache", "my-key")
CacheGetResponse(value=b'my-value, status=<CacheGetStatus.HIT: 1>)

>>> client.multi_get("my-cache", "key1", "key2")
CacheMultiGetResponse(responses=[CacheGetResponse(value=b'value1', status=<CacheGetStatus.HIT: 1>), CacheGetResponse(value=b'value2', status=<CacheGetStatus.HIT: 1>)])
```
which is much more informative and civilized.

### DIFF
--- a/src/momento/cache_operation_types.py
+++ b/src/momento/cache_operation_types.py
@@ -20,9 +20,6 @@ class CacheSetResponse:
         Args:
             key (bytes): The value of the key of item that was stored in cache..
             value (bytes): The value of item that was stored in the cache.
-
-        Raises:
-            InternalServerError: If server encountered an unknown error while trying to store the item.
         """
         self._value = value
         self._key = key
--- a/src/momento/cache_operation_types.py
+++ b/src/momento/cache_operation_types.py
@@ -149,10 +149,6 @@ class CacheDeleteResponse:
     def __init__(self) -> None:
         pass
 
-    @staticmethod
-    def from_grpc_response(grpc_create_cache_response: Any) -> "CacheDeleteResponse":  # type: ignore[misc]
-        return CacheDeleteResponse()
-
     def __str__(self) -> str:
         return self.__repr__()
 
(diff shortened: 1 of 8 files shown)


src/momento/cache_operation_types.py
```diff
@@ -130,54 +138,97 @@ def values_as_bytes(self) -> List[Optional[bytes]]:
...
         return f"CacheMultiGetResponse(responses={self._responses!r})"
 
 
 class CacheDeleteResponse:
-    def __init__(self, grpc_create_cache_response: Any):  # type: ignore[misc]
+    def __init__(self) -> None:
         pass
 
+    @staticmethod
+    def from_grpc_response(grpc_create_cache_response: Any) -> "CacheDeleteResponse":  # type: ignore[misc]
```
eaddingtonwhite, line 153: copy pasta issue on this? `grpc_create_cache_response` should be `grpc_delete_cache_response`

src/momento/cache_operation_types.py
```diff
@@ -130,54 +138,97 @@ def values_as_bytes(self) -> List[Optional[bytes]]:
...
+        return f"CreateCacheResponse()"
+
 
 class DeleteCacheResponse:
-    def __init__(self, grpc_delete_cache_response: Any):  # type: ignore[misc]
+    def __init__(self) -> None:
         pass
 
+    @staticmethod
+    def from_grpc_response(grpc_create_cache_response: Any) -> "DeleteCacheResponse":  # type: ignore[misc]
```
eaddingtonwhite, line 183: `grpc_delete_cache_response`

src/momento/cache_operation_types.py
malandis: Good catch

malandis: I opted to just delete these methods altogether. They aren't serving any purpose for now.

eaddingtonwhite (approved):

PR #115, opened 2022-06-21
feat: Add client-side implementation of exists

### LINKED ISSUE #106: add interfaces for expire and exists to incubating pkg

malandis:
Closes #106.

This PR adds a client-side implementation of `exists`. The focus is on the response object.

While other solutions only provide the user the number of keys that exist, we improve on this. In addition to using `num_exists`, users may test if all the keys exist with `all`, or by using the response object in an if expression (or truth-value testing context). We also provide methods anticipating follow-up behavior: `missing_keys`, `present_keys`, `zip_keys_and_results`. 

Examples:

(1) Test for and find missing keys:
```python
keys = ["key1", "key2", "key3", "key4"]
response = client.exists(cache_name, *keys)
if not exists_response.all():
	print(f"Expected {len(keys)} to exist but got {response.num_exists()}"
	print(f"These were missing: {response.missing_keys()}")
	print(f"These were present: {response.present_keys()}")
```

(2) Display the status of each key:
```python
for key, does_exist in client.exists(cache_name, *keys).zip_keys_and_results():
	print(f"{=key}: {=does_exist}")
```

(3) DWIM in an if-expr:
```python
if client.exists(cache_name, "key"):
	print("It's there")
else:
	print("Oh no")
```

### DIFF
--- a/src/momento/incubating/cache_operation_types.py
+++ b/src/momento/incubating/cache_operation_types.py
@@ -225,7 +225,9 @@ class CacheExistsResponse:
         Returns:
             List[Union[str, bytes]]: List of queried keys that do not exist.
         """
-        return [key for key, does_exist in zip(self._keys, self._results) if not does_exist]
+        return [
+            key for key, does_exist in zip(self._keys, self._results) if not does_exist
+        ]
 
     def present_keys(self) -> List[Union[str, bytes]]:
         """List the keys that exist.


eaddingtonwhite (approved):

cprice404:
coming into this cold - not that we should model our API off of competitors but do you have any RTFM links for pointers to how other caches expose this?

it's not an API I've used before, so will have to let it sink it for a bit before I can formulate any clear opinions on it :)  (Don't block on me obviously.)

malandis:
> coming into this cold - not that we should model our API off of competitors but do you have any RTFM links for pointers to how other caches expose this?
> 
> it's not an API I've used before, so will have to let it sink it for a bit before I can formulate any clear opinions on it :) (Don't block on me obviously.)

Followed up off-line.

src/momento/incubating/README.md
```diff
@@ -74,10 +76,69 @@ CacheDictionaryGetAllResponse(value={b'key1': b'value1', b'key2': b'value2'}, st
...
+
+```python
+>>> response = client.exists("my-cache", *keys)
+>>> response.present_keys()
+```
+
+6. Iterate over all keys
+
+```python
+>>> for key, does_exist in client.exists("my-cache", *keys).zip_keys_and_results():
```
cprice404, line 132: any reason not to just return a `dict` on this one?  callers can use language constructs to decide how they want to iterate over the results.

or maybe that is what you are returning here and python's `for` is just destructuring?  if that's the case, not sure you need the word `zip` in the fn name?

I would maybe just call this `results()` and have it return a `dict[key, bool]`.

tbh I'm not even sure if you need the `missing_keys` / `present_keys` with that approach since callers can just do a filter operation.  Those functions seem nice but i feel like it's always best to try to minimize the surface area of an API at the beginning and only add new things as people run into concrete requirements for them that can't be trivially solved in the language itself.

cprice404 (commented):
overall looks nice!  just one food for thought comment about returning a `dict` and letting callers lean into the language primitives to allow us to eliminate a few of the functions and reduce API surface area.

PR #124, opened 2022-07-19
feat: add support for retrying failed requests

cprice404:
This commit adds a retry interceptor that can retry certain
types of failed requests.  Configuration values are hard-coded
for now, until we do some basic design work for configurable
tuning.

Also includes some tweaks to logging to ensure that we don't
interpolate log message strings for messages that the logger
is configured to exclude due to the log level.


src/momento/aio/_scs_grpc_manager.py
```diff
@@ -37,18 +36,30 @@ class _DataGrpcManager:
...
+        Header("agent", f"python:{_ControlGrpcManager.version}"),
+    ]
+    interceptors = [
+        AddHeaderClientInterceptor(headers),
+        *get_retry_interceptor_if_enabled()
+    ]
+    for interceptor in interceptors:
+        print(f"CHECKING INTERCEPTOR: {interceptor} ({interceptor.__class__})")
+        print(f"IS UNARYUNARY: {isinstance(interceptor, grpc.UnaryUnaryClientInterceptor)}")
+        print(f"IS AIO UNARYUNARY: {isinstance(interceptor, grpc.aio.UnaryUnaryClientInterceptor)}")
```
kvcache, line 64: looks like maybe you forgot to rm some debuggy stuff

kvcache (commented):
direction looks good. Missed some `print`s and looks like the linter has yet to extract its tax from you

src/momento/aio/_scs_grpc_manager.py
cprice404: ty

malandis (approved):

Issue #120, opened 2022-07-14
python sdk loadgen+perf


cprice404:
Support multiple gRPC channels in python SDK and do some initial tuning to decide how many to default to.

cprice404:
I have found a handful of settings and fixes that I will put into a PR soon.  Here's a matrix of relevant perf results for a few different settings configurations.

In all of these runs we set `requestTimeoutMs` to `1000` to make sure it was easy to reproduce some percentage of timeouts.  request pattern is 1 set followed immediately by 1 get; payload is 2500 bytes.

NOTE: TPS and latencies in this table only include successful requests.

| # workers | # clients | grpc tuning applied | % success | % timeout | tps | p50    | cpu% |
|-----------|----------|--------------------|------------|-----------|-------|------|-----|
| 200          | 1              | no                            | 100%           | 0%          | 3680 | 47ms | 90% |

cprice404:
https://github.com/momentohq/oncall-tracker/issues/186

PR #125, opened 2022-07-20
feat: support for multiple gRPC channels

cprice404:
This commit is the result of several hours of experimentation
with gRPC tuning settings and running load tests with varying
numbers of gRPC channels.

See results here:
https://github.com/momentohq/client-sdk-python/issues/120

The tl;dr is:

* the python gRPC client seems to be tuned reasonably well out-of-the-box
  and I wasn't able to demonstrate any obvious perf improvements by
  using multiple channels or tweaking gRPC configs.
* the javascript SDK performs slightly better than the python one, which
  was surprising to me.  My best guess based on the data is that something
  about the async / network I/O frameworks in nodejs is able to do a tiny
  bit better job utilizing the CPU core, and that gives it a slight edge.


src/momento/aio/simple_cache_client.py
```diff
@@ -289,7 +305,15 @@ async def delete(self, cache_name: str, key: str) -> CacheDeleteResponse:
             AuthenticationError: If the provided Momento Auth Token is invalid.
             InternalServerError: If server encountered an unknown error while trying to delete the item.
         """
-        return await self._data_client.delete(cache_name, key)
+        return await self._get_next_client().delete(cache_name, key)
+
+    def _get_next_client(self) -> _ScsDataClient:
+        # print(f"Using client {self._next_client_index}")
```
eaddingtonwhite, line 311: mean to leave this? Maybe should be a debug log statement?

eaddingtonwhite (approved):
Nice perf testing write up in ticket. This looks like a good base can iterate on in future prs 👍 🚢

src/momento/aio/_scs_grpc_manager.py
```diff
@@ -40,6 +41,20 @@ def __init__(self, auth_token: str, endpoint: str):
             target=endpoint,
             credentials=grpc.ssl_channel_credentials(),
             interceptors=_interceptors(auth_token),
+            # Here is where you would pass override configuration to the underlying C gRPC layer.
+            # However, I have tried several different tuning options here and did not see any
+            # performance improvements, so sticking with the defaults for now.
```
malandis, line 46: What tuning options did you try?

src/momento/aio/simple_cache_client.py
```diff
@@ -72,7 +88,7 @@ async def __aexit__(
         traceback: Optional[TracebackType],
     ) -> None:
         await self._control_client.close()
-        await self._data_client.close()
+        await self._get_next_client().close()
```
malandis, line 91: We should close all the data clients, right?

src/momento/aio/_scs_grpc_manager.py
kvcache: Probably the ones commented below :-p

src/momento/aio/simple_cache_client.py
cprice404: yipes!  thanks for catching this

src/momento/aio/_scs_grpc_manager.py
cprice404: correct.  I also tried a few other random things like reducing the max message size to 1 byte, just to make sure that the library was actually respecting the settings.  But the ones I considered most promising / likely to yield results are the ones commented below.

src/momento/aio/simple_cache_client.py
kvcache: +1 changing the hard code client count to 2 would leak client instances here

src/momento/aio/simple_cache_client.py
cprice404: nope thanks will get rid of it

src/momento/aio/_scs_grpc_manager.py
cprice404: yeah so far in my experiments in both javascript and python, that concurrent streams setting has not made any noticeable difference at all.  anecdotally it feels like the client is honoring the server's configuration value, or the server's configuration value is preventing the client one from making any difference.  No scientific methodology behind that assertion, just anecdotal.

src/momento/aio/simple_cache_client.py
```diff
@@ -46,21 +46,37 @@
...
+    # configured to allow a max of 100 streams per connection.  In the javascript SDK, multiple
+    # clients resulted in an obvious performance improvement.  However, in the python SDK I have
+    # not yet been able to observe a clear benefit.  So for now, we are putting the plumbing in
+    # place so that we can more easily test performance with multiple connections in the future,
+    # but we are leaving the default value set to 1.
+    #
+    # We are hard-coding the value for now, because we haven't yet designed the API for
+    # users to use to configure tunables:
+    # https://github.com/momentohq/dev-eco-issue-tracker/issues/85
+    _NUM_CLIENTS = 1
```
gautamomento, line 59: Thank you for sharing the performance numbers and an amazing deep dive.

This is just a curiosity question - is there a reason why we chose to create multiple clients vs actually implementing channel pooling. 

https://grpc.io/docs/guides/performance/#general and the related issue https://github.com/grpc/grpc/issues/21386

I just want to make sure that gRPC won't do any sort of optimizations given that the channels have same arguments. If you have verified that the connections won't be reused then I guess this is a fine approach too.

cprice404: thanks for bringing this up.

The main reason I'm creating multiple clients is just that it seemed like an easier implementation path.

When you say "implementing channel pooling" - do you just mean creating more than one channel inside of the grpc_manager class and then round-robining over those instead of doing the round-robin out here?  If we have any reason to believe that there would be different performance characteristics for that approach then it would be worth trying out.  Let me know if you have thoughts.

I didn't put a ton of thought into this, but from a customer API / UX perspective, I kind of like this approach because it would allow advanced customers to experiment with creating their own pool of MomentoCache objects without having to think about or tune the number of gRPC channels that each client uses.

re: "channels have the same arguments" - yes I read that in the docs and I was concerned about it, but I wasn't sure what to think about it for 2 reasons:
1. creating multiple channels in the JS SDK very clearly made a difference so I kind of assumed the python one would handle the channel creation the same way, but that may be a faulty assumption.
2. That mention in the docs about the channel arguments talks about "using a use-specific channel arg such as channel number", but then if you look at the grpc source where they steer you to read about config flags, the only setting I could find that looked related to this was the [`channel_id`](https://github.com/grpc/grpc/blob/v1.46.x/include/grpc/impl/codegen/grpc_types.h#L445-L446) setting, and the comments in the code make it sound like that is only used for Objective-C.  I didn't see any other settings that looked like they would cover what was being hinted at by the docs you linked above.

I can try out that channel id setting pretty easily, no harm in that.

If you see another setting in that list that looks like it would be suitable for that task I would definitely be interested in your thoughts there.

And lastly if you have any experience messing with their instrumentation / census / channelz stuff and you know an easy way for me to verify whether it is optimizing out the additional channels, I'd also love to know more about that!  If not I can circle back around in the future and dig into that more.

gautamomento: I was always thinking this would be done at a lower level, so I had created this abstraction, so that we dont have to leak number of clients etc. in the higher level code as we are doing now

https://github.com/momentohq/client-sdk-python/blob/main/src/momento/aio/_scs_grpc_manager.py

cprice404: yep, that's definitely an option, and I'm not 100% certain that the approach I've taken here is better.  But it does allow the customers to construct multiple instances of the client directly if they want to, if they opt out of using `SimpleCacheClient`.  And in languages like python/JS where you really have to spawn multiple processes to utilize multiple CPU cores, since we can't share the gRPC channels across processes, I guess I was thinking it's more natural to abstract away the gRPC part by keeping the relationship 1-1 there, and allow customers to create multiple instances of the client directly if they want more control than what SimpleCacheClient gives them.

I could be persuaded that it would be a better UX to push the abstraction down a level further so I would be happy to hear more of your thoughts about it.  cc: @kvcache

src/momento/aio/simple_cache_client.py
cprice404: fixed in https://github.com/momentohq/client-sdk-python/pull/125/commits/0ce5f6d5a74b38e9578722638c4a6363caee5303

src/momento/aio/simple_cache_client.py
cprice404: fixed in https://github.com/momentohq/client-sdk-python/pull/125/commits/ed394103eccc9848fb371fc800baab9805d2c935

src/momento/aio/simple_cache_client.py
cprice404: thanks!  I'll do a quick test using that `channel_id` setting to see if it makes any difference.

eaddingtonwhite (approved):

src/momento/aio/simple_cache_client.py
cprice404: I did some runs where I set that `channel_id` field and I didn't see any difference in performance, so, putting a pin in this for now.  It may indeed be worth exploring later in the future.  Especially if we have time to dig into the gRPC instrumentation to see what kind of insights that gives us.

cprice404:
rebased onto latest `main`

eaddingtonwhite (approved):

malandis (approved):

Issue #128, opened 2022-07-28
add CI to ensure that python examples can be run from different python versions


cprice404:
This is somewhat related to https://github.com/momentohq/client-sdk-examples/issues/99 .

We need to make sure the python examples repo is easy to run and builds properly from different versions of python.  (Probably at least 3.7, 3.8, 3.9, 3.10 for now.)

In the client-sdk-python repo we are using `tox` to test against different versions of python.  We could do the same for the `client-sdk-examples` repo, but we don't want users to have to install or know anything about `tox` in order to run the examples.  We want to keep the examples simple, probably approachable via nothing but `pip`.  (They currently use `pipenv` but it'd be better if we didn't even require that.  I think that when I switched it over to `pipenv` it hard-coded python 3.9 into the lock file, and I haven't had time to test what happens if you try to consume it from a different python version.)

So, for CI we need something like tox, but however it is set up it needs to hopefully just rely on a requirements.txt or setup.py or something that will work okay with just `pip`.

Also fair game to use something besides `tox` for the CI to run stuff against the matrix of python versions, but if we do that we should probably cut over the `client-sdk-python` repo to match.

PR #131, opened 2022-08-23
Remove tox from ci

pgautier404:
Remove tox from the testing and linting process in favor of a more transparent github actions setup. This maintains consistency with the client-sdk-examples repository flow. Addresses #94 and #128.


eaddingtonwhite (dismissed):
LGTM. I didnt try locally though just FYI. I would load this into vscode and pycharm/intelij  make sure linting is still consistent between dev set ups as well.

.github/workflows/on-push-to-release-branch.yml
```diff
@@ -49,11 +75,24 @@ jobs:
...
 
-      - name: Build package and run tests
+      - name: Build package
+        run: |
+          pip install -r requirements.txt
+          python -m build
+
+      - name: Run linters
+        id: linters
+        uses: ./.github/linters
```
cprice404, line 85: same nitpick as on previous pr, it might be more obvious what that file was for if you set up the directory structure as `./.github/actions/linters`.  Not a blocker if you disagree.

.github/linters/action.yaml
```diff
@@ -0,0 +1,42 @@
...
+    description: "Black Result"
+    value: ${{ steps.set-outputs.outputs.black }}
+  flake8:
+    description: "Flake8 Result"
+    value: ${{ steps.set-outputs.outputs.flake8 }}
+runs:
+  using: "composite"
+  steps:
+      - name: Mypy linting
+        id: mypy
```
cprice404, line 16: this doesn't change the mypy config we were using before in any way, correct?  just tweaks how we are launching it?

cprice404 (dismissed):
this is great, thanks for doing it.  one nit, no blockers.

detox

.github/workflows/on-push-to-release-branch.yml
pgautier404: Done, thanks for the suggestion.

.github/linters/action.yaml
pgautier404: Yep, there's a `mypy.ini` that wasn't modified in any way.

cprice404 (approved):

malandis (changes requested):

malandis (approved):

Issue #117, opened 2022-06-22
tidy up python SDK


malandis:
- [x] Use GUIDs in tests
- [x] Use poetry to manage dependencies
- [x] Use flake8, isort
~~- [ ] Use pydantic for validation~~
- [x] Remove init functions. These are emulating builder patterns, which are not necessary in Python. Python supports optional and keyword arguments which obviate the need for builders.
- [x] Use pytest for testing

PR #132, opened 2022-09-07
chore: use guids in tests

malandis:
Use random data in integration tests. The previous code used specific strings for keys and values. Because the integration tests run repeatedly on the same cache, this could cause undesirable behavior if the same tests are repeated in quick succession.

Work towards #117


eaddingtonwhite (approved):
Looks clean 🧼

PR #134, opened 2022-09-09
refactor: use pytest

malandis:
This PR refactors our `unittest` test scripts to use `pytest` instead.

unittest has several shortcomings, chief among which was our async integration tests for python 3.7 did not work, and they did not work in a bad way. The test runner was short circuiting the async tests and reporting they all passed. Upon investigation, unittest for python 3.7 does not support async test cases. To get around this, we were backporting code from python 3.8 to run async tests in python 3.7. This did not work as intended. Hence a more pressing need to switch to pytest.

Pytest not only does not suffer this deficiency, but has many upsides. It has a rich plugin ecosystem and a better test fixture framework. The tests are functional in style -- as opposed to jUnit -- and have far less boilerplate.

Also in this PR: since we have both an async client and a synchronous client, we had integration tests for both. Because the synchronous client just delegates behavior to the async client, the tests were identical modulo `async`/`await` keywords. The tests _are_ necessary though to check for mistakes in the code that delegates requests. To better manage this, we introduce a script to port the async tests to the sync client. That way we only need to develop tests in one place.

Work towards #117


tests/scripts/sync_from_async.sh
```diff
@@ -0,0 +1,34 @@
...
+    async_file=$1
+    sync_file=$2
+
+    cp $async_file $sync_file
+
+    sed -i -e 's/async //g' $sync_file
+    sed -i -e 's/await //g' $sync_file
+    sed -i -e 's/momento.aio./momento./g' $sync_file
+    sed -i -e 's/momento.incubating.aio.simple/momento.incubating.simple/g' $sync_file
+    sed -i -e 's/client_async/client/g' $sync_file
```
cprice404, line 25: this seems like it could be fraglie, but we can give it a try.

cprice404 (approved):
thanks for investigating the `3.7` issue!

tests/scripts/sync_from_async.sh
malandis: Agree. I timeboxed how to write a single set of tests that could be re-used between the two clients and be readable. This was a stop gap until we can find a more graceful approach.

PR #137, opened 2022-09-14
refactor: remove init functions

malandis:
The `init` functions to build the cache clients are unnecessary. Because python allows keyword arguments, we can pass any and all configuration options in the constructor.

Work towards #117


bruuuuuuuce (approved):
:shipit:

PR #139, opened 2022-09-16
chore: move Python SDK examples from `client-sdk-examples` repo

poppoerika:
Also added dependabot and updated the `on-pull-request` workflow.


poppoerika:
@cprice404 @malandis @pgautier404 
It looks like we use `pipenv` for the examples but `poetry` for the client-sdk-python. 
Based on the dev eco public launch readiness, this can be updated later?

malandis (approved):
:confetti_ball:

malandis:
> @cprice404 @malandis @pgautier404 It looks like we use `pipenv` for the examples but `poetry` for the client-sdk-python. Based on the dev eco public launch readiness, this can be updated later?

Yes we can sort that out later.

PR #163, opened 2022-12-22
feat: migrate synchronous API to consume gRPC synchronous API

cprice404:
Prior to this commit, our synchronous API was implemented as a wrapper around our asynchronous API, with the intention of reducing code duplication and making it easier to maintain both.

However, after deeper investigation of how that
approach to the synchronous API implementation behaves in applications that are using asyncio for other purposes, we have decided that it will be safer and a better user experience to implement it such that it directly consumes the gRPC synchronous API.

This may require a bit more code and maintenance, but it will give us the best UX for users who need to consume our synchronous API in an app that uses other asyncio libraries.


malandis:
Will DRY code in future PRs.

malandis (approved):

PR #165, opened 2022-12-28
fix: pin ubuntu and poetry ci dependencies

malandis:
Pins the ubuntu version to 20.04. Previously we used `ubuntu-latest`; GitHub started pointing this from 20.04 to 22.04 in October and the integration tests started failing. We have timeboxed an investigation into why 22.04 fails, but remains inconclusive. Hence in the meantime we revert to 20.04. We also pin the poetry dependency as a matter of good practice.


pgautier404 (approved):
:shipit:

PR #169, opened 2023-01-05
refactor: Reduce pass-through args in BasicPythonLoadGen.

schwern:
The load-gen script passes through a lot of individual variables which makes it unwieldy to change its options. Instead, make them available as BasicPythonLoadGen attributes.

* Group config variables in BasicPythonLoadGenOptions.
  * Make it an BasicPythonLoadGen attribute.
* Make BasicPythonLoadGenContext a BasicPythonLoadGen attribute.

### DIFF
--- a/examples/example_load_gen.py
+++ b/examples/example_load_gen.py
@@ -47,6 +47,7 @@ class BasicPythonLoadGenOptions:
     show_stats_interval_seconds: int
     total_seconds_to_run: int
 
+
 @dataclass
 class BasicPythonLoadGenContext:
     start_time: float
@@ -285,9 +286,7 @@ If you have questions or need help experimenting further, please reach out to us
 """
 
 
-async def main(
-    options: BasicPythonLoadGenOptions
-) -> None:
+async def main(options: BasicPythonLoadGenOptions) -> None:
     initialize_logging(options.log_level)
     load_generator = BasicPythonLoadGen(options)
     await load_generator.run()
--- a/examples/example_load_gen.py
+++ b/examples/example_load_gen.py
@@ -38,6 +38,16 @@ class AsyncSetGetResult(Enum):
 
 
 @dataclass
+class BasicPythonLoadGenOptions:
+    log_level: int
+    request_timeout_ms: int
+    cache_item_payload_bytes: int
+    max_requests_per_second: int
+    number_of_concurrent_requests: int
+    show_stats_interval_seconds: int
+    total_seconds_to_run: int
+
+@dataclass
 class BasicPythonLoadGenContext:
     start_time: float
     get_latencies: HdrHistogram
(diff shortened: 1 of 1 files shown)


examples/example_load_gen.py
```diff
@@ -37,6 +37,16 @@ class AsyncSetGetResult(Enum):
     THROTTLE = ("THROTTLE",)
 
 
+@dataclass
+class BasicPythonLoadGenOptions:
```
cprice404, line 41: I love this.

examples/example_load_gen.py
```diff
@@ -55,38 +65,26 @@ class BasicPythonLoadGenContext:
...
-            self.auth_token, cache_item_ttl_seconds, self.request_timeout_ms
+            self.auth_token, cache_item_ttl_seconds, self.options.request_timeout_ms
         ) as cache_client:
             try:
                 await cache_client.create_cache(BasicPythonLoadGen.cache_name)
             except momento.errors.AlreadyExistsError:
                 self.logger.info(f"Cache with name: {BasicPythonLoadGen.cache_name} already exists.")
 
-            load_gen_context = BasicPythonLoadGenContext(
+            self.context = BasicPythonLoadGenContext(
```
cprice404, line 87: i'm a little more iffy about this.  i really don't like having member variables that contain mutable state in async/concurrent contexts because it can make the coordination semantics more difficult (and slow).

In python this is a non-issue because of the GIL so I'm okay with this if you prefer it.  But in the languages that support parallelism we will probably want to scrutinize this sort of change more thoroughly.  Making the functions as pure as possible (not relying on side effects / object state, and preferring arguments) can really make a huge difference in code maintainability.

cprice404 (approved):

examples/example_load_gen.py
schwern: Yes, there's no threading here to worry about. And if there was we'd need to make sure they're all using the same BasicPythonLoadGenContext regardless.

If there was, I'd start by extracting several classes out of BasicPythonLoadGen; it does too much.

cprice404: > If there was, I'd start by extracting several classes out of BasicPythonLoadGen; it does too much.

Yes, if it was production code I would not have designed it this way.  The intent is for users to be able to easily play with it and modify it if they like, to test different operations or payload types.  That is why it is one class in one file.

cprice404: > Yes, there's no threading here to worry about. And if there was we'd need to make sure they're all using the same BasicPythonLoadGenContext regardless.

Not necessarily.  I would probably have each worker keep track of its own state and then have the main thread aggregate them so that there was no lock contention or other issues related to shared mutable state to deal with.  So I would very likely be trying hard to avoid storing the mutable state in a widely accessible class variable.

But this is partially a stylistic consideration, and again I acknowledge that it is not relevant for this particular Python environment so I am okay with the change if you prefer it.  I'm just trying to give you a heads up on my thoughts as we move into .NET, Java, go, Rust, etc.

malandis (approved):

PR #171, opened 2023-01-06
fix: re-introduce retry interceptor for synchronous client

cprice404:
In an earlier commit we made some changes to the implementation
of the synchronous client to use the grpc synchronous API directly,
rather than wrapping the asyncio API, which was causing problems
in environments that used other asyncio libraries.

During the rollout of that change we mistakenly lost the retry
capabilities that we have in the asyncio client.  This commit
adds the retry suppport back in for the sync client.

This implementation is not DRY and we need to clean it up as we
are doing the work to introduce Configuration, but this will
at least unblock users for now.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -57,6 +57,9 @@ disallow_any_expr           = False
 [mypy-momento.internal.synchronous._add_header_client_interceptor]
 disallow_any_expr           = False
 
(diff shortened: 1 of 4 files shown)


cprice404:
This should address https://github.com/momentohq/client-sdk-python/issues/167

pgautier404 (approved):

PR #183, opened 2023-01-21
chore: organize project

### LINKED ISSUE #177: clean up file / directory structure
i believe we still have some files whose names are prefixed with `_` in the root dir; my gut is that we should move these into the `internal` subdir.  As an end user I would somewhat expect that if I pulled the source code the only files in the root dir would be public APIs.

Also the aio dir predates the `internal` dir but everything in there should probably be moved over into `internal`.

We should also just take a general pass and see if there are any other re-organizations / cleanups that we want to take care of before 1.0.

malandis:
In a previous PR we moved `aio` under `internal` so that it parallels `synchronous` and `common`. We also moved the cache client class modules directly under the `momento` package.

Across the board we take advantage of the init files to roll up classes and functions that are nested under various packages and modules. The principle here is to make commonly accessed things easy to access. Thus a user now needs to write:

```python
# 1. Access the clients
from momento import SimpleCacheClient, SimpleCacheClientAsync

# instead of before

from momento.simple_cache_client import SimpleCacheClient
from momento.aio.simple_cache_client import SimpleCacheClient

# 2. Access the responses
from momento.responses import CacheGet, CacheSet

# Access error types
from momento.errors import SdkException, InvalidArgumentException

# Access auth helpers
from momento.auth import EnvMomentoTokenProvider

# Access configs
from momento.config import Laptop
```

In this particular PR we made the following changes:

- chore: move _utilities under internal
- chore: roll up configuration classes under config package init
- chore: roll up auth classes under auth package

Closes #177


cprice404 (approved):
awesome cleanup.  Thanks!
approve

PR #191, opened 2023-01-26
fix: Ensure test caches are cleaned up.

### LINKED ISSUE #62: Delete integration test caches on test failure
Currently integration test caches are created on every run, and they aren't deleted when tests fail. Over time, our internal alpha environment will accumulate a bunch of orphaned caches. We should add a `finally` type clause to the tests so even when failures occur the test caches are deleted at the end of the run.

schwern:
* Even if the test fails.
* And for caches with unique names.
* Harden the default cache name against parallel matrix test runs and parallel PRs.

Now that we're deleting the cache per individual test, the parallel Python version tests collided.

Closes #62


cprice404 (dismissed):
salute

cprice404 (approved):

PR #193, opened 2023-01-26
test: Use shared tests for the scalar tests. Switch to pytest-describe style.

### LINKED ISSUE #189: Make a shared test for validating cache names.
This is both a good idea, everything needs to validate the cache names, and also to figure out the pattern of writing shared tests in PyTest. Do it for the scalar tests.

schwern:
* Establishes a pattern of shared tests for the coming collections expansion.
* Organizes the tests [pytest-describe](https://github.com/pytest-dev/pytest-describe) style by method.

This just moves tests around, no test was removed. It might be easier to review commit-by-commit with whitespace changes off.

Closes #189

### DIFF
--- a/tests/momento/simple_cache_client/test_scalar_async.py
+++ b/tests/momento/simple_cache_client/test_scalar_async.py
@@ -1,5 +1,7 @@
 import time
 from datetime import timedelta
+from functools import partial
+from typing import Awaitable, Callable, Union
 
 from pytest import fixture
 from pytest_describe import behaves_like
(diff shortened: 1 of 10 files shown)


tests/momento/simple_cache_client/test_scalar_async.py
```diff
@@ -9,274 +12,268 @@
...
+
+@behaves_like(a_cache_name_validator)
+@behaves_like(a_key_validator)
+@behaves_like(a_connection_validator)
+def describe_get():
+    @fixture
+    def cache_name_validator(client_async: SimpleCacheClientAsync):
+        def _cache_name_validator(cache_name):
+            key = uuid_str()
+            return client_async.get(cache_name, key)
```
malandis, line 118: At first I was unsure how we're invoking an async method (`client_async.get`) without the outer functions annotated as `async`. I suppose this is returning a coroutine object that is not on the event loop, and that is awaited in the caller functions above?

tests/momento/simple_cache_client/test_scalar_async.py
```diff
@@ -9,274 +12,268 @@
...
+        client_async: SimpleCacheClientAsync, cache_name_validator
+    ) -> None:
+        response = await cache_name_validator("")
+        assert response.error_code == MomentoErrorCode.INVALID_ARGUMENT_ERROR
+        assert response.inner_exception.message == "Cache header is empty"
+
+    async def with_bad_cache_name_throws_exception(client_async: SimpleCacheClientAsync, cache_name_validator) -> None:
+        response = await cache_name_validator(1)
+        assert response.error_code == MomentoErrorCode.INVALID_ARGUMENT_ERROR
+        assert response.inner_exception.message == "Cache name must be a non-empty string"
```
malandis, line 40: I was going to enable mypy checks on the tests shortly. The functions in this file (and the other) will need return types for all the functions. Most of them return None, though others will return the response types.

tests/momento/simple_cache_client/test_scalar_async.py
```diff
@@ -9,274 +12,268 @@
...
+
+
+def a_key_validator():
+    async def with_null_key_throws_exception(
+        client_async: SimpleCacheClientAsync, cache_name: str, key_validator
```
malandis, line 45: We would also need annotations for the various validators. That may prove tricky.

malandis (commented):
I like this style of tests. It's a great improvement. I left a couple comments about the type hints. At a minimum we should annotate the return values. We may defer annotating the validators unless that proves a quick fix.

tests/momento/simple_cache_client/test_scalar_async.py
schwern: This is where the lack of a common response class hierarchy bites us. I think we'll be good enough by changing `class CacheGetResponse(ABC):` to `class CacheGetResponse(Response):` and `class Response(ABC)`. Then these would return Response.

tests/momento/simple_cache_client/test_scalar_async.py
schwern: This PR is just moving existing functions around, so we'll have type problems one way or another. Can we handle that in a different PR?

schwern: I'll see what I can puzzle out.

schwern: Using `key_validator` as an example, `key_validator` is returning a `<function describe_get.<locals>.key_validator.<locals>._key_validator at 0x10df21950>` and the internal `_key_validator` is returning a `<coroutine object SimpleCacheClientAsync.get at 0x10df25200>`.

I have no idea what that signature might be.

schwern: This is over my head. I'd like to get this in and then work on the types.

cprice404:
I'll defer to @malandis on this.

cprice404 (commented):
wfm when @malandis is happy

malandis:
> wfm when @malandis is happy

@cprice404 I've made some modifications and am good to go on this. Need your review now since I've contributed :)

schwern:
@cprice404 I want to do one more thing.

schwern:
@cprice404 Ok, ready.

malandis (approved):
:sparkles:

PR #197, opened 2023-01-27
chore: update examples to use latest python client

malandis:
We update the momento dependency to 0.19.0 and update the examples as
appropriate. Because pattern matching is available for 3.10+, we
provide examples with and without pattern matching.

Hence we now have examples for `{async, sync} x {pattern matching,
legacy-if-else}`


malandis:
TODO: will have to update Nihongo README

examples/README.md
```diff
@@ -12,23 +12,34 @@ _Read this in other languages_: [日本語](README.ja.md)
...
+
 ```bash
 xcode-select --install
 ```
 
+## To Pattern Match or Not to Pattern Match?
+
+Python introduced [structural pattern matching](https://peps.python.org/pep-0636/) in version 3.10. We believe your experience with our client will really shine with this feature. To demonstrate this, we provide examples with and without pattern matching. The example filenames without pattern matching are suffixed with `_prepy310`. For example, `example_async.py` demonstrates our asynchronous client _with_ pattern matching; `example_async_prepy310.py` demonstrates our asynchronous client _without_ pattern matching.
+
+If you are new to this feature, we invite you to try it. Nevertheless you do _have_ to use it. Our client is perfectly compatible with the legacy `if-elif-else` type checking with `isinstance`.
```
cprice404, line 38: This is great but I think it belongs in the main README in addition to or instead of here?  Check out the "Response Types" sections in the JS/.NET READMEs

cprice404 (approved):
this is a big step toward the finish line, ty!

examples/README.md
malandis: I put the blurb in the main README.

pgautier404 (approved):
LGTM!

cprice404 (approved):

PR #199, opened 2023-01-27
refactor: Extract internal scalar op support functions.

### LINKED ISSUE #198: Extract response functions by data type
Right now the internal code for requests and responses is in on file `src/momento/internal/common/_data_client_ops.py`. This will not scale. Extract the scalar functions into their own file.

- [x] Extract the construct response and prepare request functions into `src/momento/internal/common/_data_client_ops.py` `src/momento/internal/common/_data_client_scalar_ops.py`.
- [x] Leave the error handling wrappers alone.

schwern:
This will avoid a monolithic file and conflicts as we add collections. Each collection has its own file.

Closes #198

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -63,6 +63,9 @@ disallow_any_expr           = False
 [mypy-momento.internal.common._data_client_ops]
 disallow_any_expr           = False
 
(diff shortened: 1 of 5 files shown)


malandis (approved):
:+1:

PR #200, opened 2023-01-27
refactor: Add a CacheResponse abstract superclass for all responses.

schwern:
This lets us get the type checks right for shared behaviors of all methods.

### DIFF
--- a/src/momento/responses/response.py
+++ b/src/momento/responses/response.py
@@ -1,0 +1,5 @@
+from abc import ABC
+
+
+class CacheResponse(ABC):
+    """Parent of all responses"""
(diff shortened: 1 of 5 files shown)


src/momento/responses/response.py
```diff
@@ -0,0 +1,5 @@
+from abc import ABC
+
+
+class CacheResponse(ABC):
```
malandis, line 4: Can you add `CacheResponse` to the init file under `src/momento/responses`? That will allow us type `from momento.responses import CacheResponse` instead of `from momento.responses.response import CacheResponse`. 

Other than that looks good

malandis (commented):
One small change otherwise looks good

malandis (approved):

PR #201, opened 2023-01-28
chore: add comments as fenceposts to reduce merge conflicts

malandis:
Since we will be working concurrently on data structure methods, we
should reduce the likelihood of merge conflicts. By grouping the
methods under these comments we will make that less likely.


schwern (approved):
Very thoughtful.

PR #202, opened 2023-01-28
fix: remove extraneous echo from make target

malandis:
The echo in this target was a placeholder and produces an error. We
remove it.


schwern (approved):
Perl will add `$(NOECHO) $(NOOP)` to empty targets which usually expands to `@true` because it has to work on a million different make variants and some of them don't like empty targets. Maybe that's what was going on here.

Issue #184, opened 2023-01-23
Prepare for collections.


schwern:
Lessons learned from Javascript. Some of these may already be done.

- [x] Split integration tests into get-set-delete, cache, dictionary, list, set.
  - [x] Extract integration test setup.
- [x] Write basic shared tests to exercise the technique.
  - [x] Cache name verification.
- [x] DRY up response classes.
  - [x] Make sure there is a basic class hierarchy (Response, Error, Success, Miss, Hit) useful for sharing tests.
- [x] Write a couple collections ops to exercise the code and see patterns.
- [x] CollectionTTL

Issue #174, opened 2023-01-09
add list data structure APIs to python SDK


cprice404:
Referencing https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083#9eb4f2ebdae44a19923e9142ebdd98ec

- [x] listConcatenateBack
- [x] listConcatenateFront
- [x] listPushBack
- [x] listPushFront
- [x] listFetch
- [x] listLength
- [x] listPopBack
- [x] listPopFront
- [x] listRemoveValue

PR #205, opened 2023-01-31
feat: First collections methods.

schwern:
This adds the minimum of collections to exercise how we add them going forward and avoid colliding when working in parallel. listFetch and listConcatenateBack.

**Note**: all the duplicated synchronous files can be ignored for review purpose, they're all generated from the async code.

* Upgrade wire types to one that supports collections and new scalar methods.
  * Unpin the dependency to be a good citizen.
* Add list types
* Add some helpful collections fixtures
* Add cache name type (didn't use it everywhere)
* Add some helpful trace logging wrappers.

For #174 and #184

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ exclude = ["src/momento/internal/codegen.py"]
 [tool.poetry.dependencies]
 python = ">=3.7,<3.12"
 
-momento-wire-types = "0.31.1"
+momento-wire-types = "^0.39"
 grpcio = "^1.50.0"
 # note if you bump this presigned url test need be updated
 pyjwt = "^2.4.0"
--- a/src/momento/internal/aio/_scs_data_client.py
+++ b/src/momento/internal/aio/_scs_data_client.py
@@ -7,6 +7,7 @@ from momento_wire_types.cacheclient_pb2 import (
     _DeleteResponse,
     _GetRequest,
     _GetResponse,
+    _ListFetchRequest,
     _SetRequest,
     _SetResponse,
 )
@@ -15,7 +16,13 @@ from momento_wire_types.cacheclient_pb2_grpc import ScsStub
 from momento import logs
 from momento.auth import CredentialProvider
 from momento.config import Configuration
-from momento.internal._utilities import _validate_ttl
+from momento.errors import convert_error
+from momento.internal._utilities import (
+    _as_bytes,
+    _validate_cache_name,
+    _validate_list_name,
+    _validate_ttl,
+)
 from momento.internal.aio._scs_grpc_manager import _DataGrpcManager
 from momento.internal.aio._utilities import make_metadata
 from momento.internal.common._data_client_ops import (
@@ -35,6 +42,8 @@ from momento.responses import (
     CacheDeleteResponse,
     CacheGet,
     CacheGetResponse,
+    CacheListFetch,
+    CacheListFetchResponse,
     CacheSet,
     CacheSetResponse,
 )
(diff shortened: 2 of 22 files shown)


pyproject.toml
```diff
@@ -30,7 +30,7 @@ exclude = ["src/momento/internal/codegen.py"]
 [tool.poetry.dependencies]
 python = ">=3.7,<3.12"
 
-momento-wire-types = "0.31.1"
+momento-wire-types = "^0.39"
```
malandis, line 33: Let's pin the version

```suggestion
momento-wire-types = "0.39.0"
```

src/momento/internal/aio/_scs_data_client.py
```diff
@@ -135,7 +143,25 @@ async def execute_delete_request_fn(req: _DeleteRequest) -> _DeleteResponse:
 
     # DICTIONARY COLLECTION METHODS
 
-    # LIST COLLECTION METHODS
```
malandis, line 138: I think we may get merge conflicts unless we keep the line above

malandis (commented):
So far so good :+1:

pyproject.toml
schwern: Why's that? Wire types should be backwards compatible.

malandis: That's true they should be. Pinning the version is just a black and white best practice so that what we test in CI is what users install.

malandis: That's a great point. For the wire types probably. We should also reevaluate how we pinned grpcio as that is likely to be used by popular Google libraries.

src/momento/internal/aio/_scs_data_client.py
schwern: Whether we think of collections as "not existing" vs "being empty" is something we need to document for the user. I guess we implicitly destroy an empty collection, so "not existing". I think we need a little doc section explaining that.

src/momento/responses/list_data/list_concatenate_back.py
```diff
@@ -0,0 +1,41 @@
...
+from dataclasses import dataclass
+
+from momento.errors import SdkException
+
+from ..mixins import ErrorResponseMixin
+from ..response import CacheResponse
+
+
+class CacheListConcatenateBackResponse(CacheResponse):
+    """Response type for a list_concatenate_back request. Its subtypes are:
```
malandis, line 11:
```suggestion
    """Response type for a `list_concatenate_back request`. Its subtypes are:
```

src/momento/responses/list_data/list_fetch.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import List
+
+from momento.errors import SdkException
+
+from ..mixins import ErrorResponseMixin
+from ..response import CacheResponse
+
+
+class CacheListFetchResponse(CacheResponse):
+    """Response type for a list fetch request. Its subtypes are:
```
malandis, line 12:
```suggestion
    """Response type for a `list_fetch` request. Its subtypes are:
```

tests/momento/simple_cache_client/test_list_async.py
```diff
@@ -0,0 +1,157 @@
...
+from tests.utils import uuid_str
+
+from .shared_behaviors_async import (
+    TCacheNameValidator,
+    TConnectionValidator,
+    a_cache_name_validator,
+    a_connection_validator,
+)
+
+TListNameValidator = Callable[[str], Awaitable[CacheResponse]]
```
malandis, line 21:
```suggestion
TListNameValidator = Callable[[TListName], Awaitable[CacheResponse]]
```

malandis (commented):
Couple small changes.

Also it looks like [tests/momento/internal/test_cache_list_fetch.py](https://github.com/momentohq/client-sdk-python/pull/205/files#diff-6d814417fb4a74c984198f19a8d2058eff944cf03918fcebcb3c315cc5928a04) is unused?

src/momento/responses/list_data/list_concatenate_back.py
schwern: If that's the style, I'll retrofit the scalar docs, too.

malandis (approved):

PR #206, opened 2023-01-31
feat: export types to be PEP561 compliant

malandis:
PEP561 mandates that, for "package mantainers who wish to support type
checking of their code MUST add a marker file named `py.typed` to
their package supporting typing".

Previously we did not have this file. That means that code importing momento
 will get errors from mypy about missing momento type imports.

See [here](https://peps.python.org/pep-0561/) for more details.


schwern (approved):
Do we need to concern ourselves with distutils?

> To have this file installed with the package, maintainers can use existing packaging options such as package_data in distutils, shown below.

malandis:
> Do we need to concern ourselves with distutils?
> 
> > To have this file installed with the package, maintainers can use existing packaging options such as package_data in distutils, shown below.

I verified poetry does the right thing -- I ran `poetry build` to generate the package distributable, then installed that in a new project.

PR #207, opened 2023-01-31
fix: Unpin dependencies to avoid library conflicts.

schwern:
Pinned dependencies invite conflicts with other libraries with the same dependencies.

Keep the minimum version, but relax the max. Stay within the major version.


nand4011 (approved):

schwern:
@eaddingtonwhite

PR #209, opened 2023-01-31
docs: Move explaining how to use responses into SimpleCacheClient.

schwern:
Cutting and pasting these docs will become untenable as the interface expands. Document the pattern in SimpleCacheClient and refer back to it.

### DIFF
--- a/src/momento/simple_cache_client.py
+++ b/src/momento/simple_cache_client.py
@@ -42,7 +42,32 @@ from momento.responses import (
 
 
 class SimpleCacheClient:
-    """Synchronous Simple Cache Client"""
+    """Synchronous Simple Cache Client
+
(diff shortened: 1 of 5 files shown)


src/momento/simple_cache_client.py
```diff
@@ -42,7 +42,32 @@
...
+    or equivalently in earlier versions of python:
+
+        response = client.delete(cache_name, key)
+        if isinstance(response, CacheDelete.Success):
+            ...
+        elif isinstance(response, CacheDelete.Error):
+            ...
+        else:
+            raise Exception("This should never happen")
+    """
```
malandis, line 70: Should add this in the async client too

cprice404 (approved):

cprice404 (approved):

PR #210, opened 2023-02-01
feat: add signing key control operations back in

malandis:
We partially revert 60df3d to add the signing key control operations
back. The `MomentoSigner` class remains in the other repository.


cprice404 (approved):
bed goes up

pgautier404 (approved):

PR #213, opened 2023-02-01
chore: add module centralizing type hints

malandis:


schwern (approved):

src/momento/typing.py
```diff
@@ -0,0 +1,14 @@
+from typing import Union
+
+# Scalar Types
+TScalarKey = Union[str, bytes]
+TScalarValue = Union[str, bytes]
+
+# Collections
+TCollectionKey = str
```
schwern, line 8: The term used by collections is name. TCollectionKey and TScalarKey are different things.

```suggestion
TCollectionName = str
```

schwern (changes requested):

schwern (approved):

PR #214, opened 2023-02-01
fix: isolate test cache names on push to release

malandis:
The CI caches clobber each other on release because all python
versions share the same cache name. We create a unique cache name
derived from the python version and commit hash as in the
`on-pull-request` workflow.


schwern (approved):
I'll put in an issue to de-dup the test jobs.

PR #218, opened 2023-02-01
fix: update example location

malandis:
We have updated the examples and forked based on python version. In
the previous example update we referenced the py310 synchronous
example as the canonical one. We update that here as well.


malandis:
TODO: it may make sense to bake in an "alternate example" to the README generation. That way we could reference both the py310 and the prepy310 examples.

cprice404 (approved):

cprice404:
> TODO: it may make sense to bake in an "alternate example" to the README generation. That way we could reference both the py310 and the prepy310 examples.

I wholeheartedly agree with this, feel free to create a ticket if you like

PR #224, opened 2023-02-01
feat: Validate the CollectionTtl.ttl is positive.

### LINKED ISSUE #222: Validate CollectionTtl
Do this in the CollectionTtl initializer, not in each method. A benefit of using an object is we can make it impossible to be invalid.

- [ ] raise ValueError if the ttl is negative

schwern:
We do this inside CollectionTtl so we don't have to check it in every collection method.

I excluded 0 because why would you want that? We can allow it later if someone really needs it.

Closes #222

### DIFF
--- a/src/momento/requests/collection_ttl.py
+++ b/src/momento/requests/collection_ttl.py
@@ -18,6 +18,9 @@ class CollectionTtl:
     The default behavior is to refresh the TTL (to prolong the life of the
     collection) each time it is written.  This behavior can be modified
     by calling the method `with_no_refresh_ttl_on_updates`.
+
+    Raises:
+        ValueError: if the ttl is not a positive amount of time.
     """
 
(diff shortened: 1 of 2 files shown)


src/momento/requests/collection_ttl.py
```diff
@@ -30,6 +33,10 @@ class CollectionTtl:
     initially created.
     """
 
+    def __post_init__(self) -> None:
+        if self.ttl is not None and self.ttl.total_seconds() <= 0:
+            raise ValueError("the TTL must be a positive amount of time")
+
```
malandis, line 39: Cool use of `__post_init__`

malandis (approved):

src/momento/requests/collection_ttl.py
schwern: Stack Overflow FTW

PR #225, opened 2023-02-01
chore: update codegen to handle examples in docstrings

malandis:
Add more regex replacements. Refactor `SimpleString` replacements into
separate data driven class.


schwern (approved):
From the tests, this looks like it should do it!

PR #226, opened 2023-02-02
chore: refactor credential provider, add string provider

### LINKED ISSUE #223: Implement StringMomentoTokenProvider
See https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083#5205ec163d14423e82e1ea472be7b83e

CC @anp13

malandis:
Make credential provider pythonic. Add the equivalent of StringMomentoAuthProvider as a static constructor.

Closes #223

### DIFF
--- a/src/momento/auth/__init__.py
+++ b/src/momento/auth/__init__.py
@@ -1,1 +1,1 @@
-from .credential_provider import CredentialProvider, EnvMomentoTokenProvider
+from .credential_provider import CredentialProvider
--- a/src/momento/auth/credential_provider.py
+++ b/src/momento/auth/credential_provider.py
@@ -1,49 +1,47 @@
+from __future__ import annotations
+
 import os
-from abc import ABC, abstractmethod
+from dataclasses import dataclass
 from typing import Optional
 
 from . import momento_endpoint_resolver
 
 
-class CredentialProvider(ABC):
+@dataclass
+class CredentialProvider:
     """Provides information that the SimpleCacheClient needs in order to establish a connection to and authenticate with
     the Momento service.
     """
 
-    @abstractmethod
-    def get_auth_token(self) -> str:
-        pass
+    auth_token: str
+    control_endpoint: str
+    cache_endpoint: str
 
-    @abstractmethod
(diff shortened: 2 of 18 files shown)


cprice404:
3 comments on this:

1. If it is blocking canary work, then this is fine and let's merge it and deal with my other comments as fast-follow.
2. Do we feel strongly that the "pythonic" UX of this will make such a big difference to our users that we are justified in diverging from how this looks/feels in other languages?
3. If we do feel the UX improvement is that significant, couldn't we still keep the original class-based providers (one for string, one for env), and then just add some static factory function that looks and feels like what you've done here but simply delegates to the constructors of those classes?  That way we could retain the same shape and feel that the other SDKs have.  I am also expecting that when SSO comes in we are going to end up with more implementations of this that may be more complex than the current ones, so I would prefer to keep the ability to separate them cleanly by class.

Open to discussion on all of the points above except (1) :)

cprice404 (commented):
this is fine to move forward with if canary is blocked but i left some questions i'd like to at least consider for fast follow

schwern:
@cprice404 

>     2. Do we feel strongly that the "pythonic" UX of this will make such a big difference to our users that we are justified in diverging from how this looks/feels in other languages?

I think so. Everything else using properties. SDK users are unlikely to be using SimpleCacheClient in multiple languages.

>     3. If we do feel the UX improvement is that significant, couldn't we still keep the original class-based providers (one for string, one for env), and then just add some static factory function that looks and feels like what you've done here but simply delegates to the constructors of those classes?  That way we could retain the same shape and feel that the other SDKs have.  I am also expecting that when SSO comes in we are going to end up with more implementations of this that may be more complex than the current ones, so I would prefer to keep the ability to separate them cleanly by class.

I think we should change the S&P to this one. It really is one class with two constructors. We can provide EnvMomentoCredentialProvider and StringMomentoCredentialProvider wrappers were they've already been published.

schwern:
@cprice404 Fast-follow: #233

cprice404 (approved):

PR #231, opened 2023-02-02
fix: Copy doc changes to SimpleCacheClient.

schwern:
I forgot to `make gen-sync` before pushing when I changed the SimpleCacheClientAsync dos.

### DIFF
--- a/src/momento/simple_cache_client.py
+++ b/src/momento/simple_cache_client.py
@@ -135,12 +135,6 @@ class SimpleCacheClient:
 
         Returns:
             CreateCacheResponse:
-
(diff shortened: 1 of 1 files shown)


cprice404 (commented):
is this supposed to be removing docs?  if it is that's fine, just not sure I know what I'm looking at here.

schwern:
> is this supposed to be removing docs? if it is that's fine, just not sure I know what I'm looking at here.

Yes, the docs were already removed from SimpleCacheClientAsync but weren't copied to SimpleCacheClient.

schwern:
Dict PR already did this.

PR #234, opened 2023-02-02
chore: migrate mypy config to pyproject.toml and simplify

malandis:
The previous mypy config has many repetitious overrides. We group
these together and put in the pyproject.toml file which centralizes
the tool configs (except flake8, which doesn't play nice).


pyproject.toml
```diff
@@ -57,6 +57,71 @@ libcst = "^0.4.9"
...
+
+disallow_any_decorated      = true
+disallow_any_explicit       = true
+disallow_any_expr           = true
+disallow_any_generics       = true
+disallow_any_unimported     = false
+disallow_subclassing_any    = false
+disallow_untyped_calls      = true
+disallow_untyped_decorators = true
+strict_equality             = true
```
schwern, line 88: Is that comment accurate? Looks like most of them are enabled.

malandis: I copied that from mypy.ini but it looks out of date.

schwern (approved):

PR #235, opened 2023-02-02
feat: Set ops

schwern:


src/momento/responses/set_data/set_remove_elements.py
```diff
@@ -0,0 +1,38 @@
...
+from dataclasses import dataclass
+
+from momento.errors import SdkException
+
+from ..mixins import ErrorResponseMixin
+from ..response import CacheResponse
+
+
+class CacheSetRemoveElementsResponse(CacheResponse):
+    """Parent response type for a `set_remove_element` request. Its subtypes are:
```
malandis, line 11:
```suggestion
    """Parent response type for a `set_remove_elements` request. Its subtypes are:
```

src/momento/responses/set_data/set_fetch.py
```diff
@@ -0,0 +1,61 @@
...
+
+        @property
+        def values_string(self) -> TSetElementsOutputStr:
+            """The values for the fetched list, as utf-8 encoded strings.
+
+            Returns:
+                TListValuesStr
+            """
+
+            return [v.decode("utf-8") for v in self.values_bytes]
```
malandis, line 44: Python has a native set type called `set` (pre 3.8 type hints: `Set`). Can instantiate with a set comrephension:

```python
return {v.decode("utf-8") for v in self.value_bytes}
```

src/momento/typing.py
```diff
@@ -31,3 +31,8 @@
 TLIstValuesOutput = Union[TListValuesOutputBytes, TListValuesOutputStr]
 
 # Set Types
+TSetName = TCollectionName
+TSetElement = Union[str, bytes]
+TSetElementsInput = Iterable[TSetElement]
+TSetElementsOutputStr = List[str]
+TSetElementsOutputBytes = List[bytes]
```
malandis, line 38:
```suggestion
TSetElementsOutputStr = Set[str]
TSetElementsOutputBytes = Set[bytes]
```

pgautier404 (dismissed):
LGTM!

src/momento/internal/_utilities/_data_validation.py
```diff
@@ -43,6 +53,14 @@ def _list_as_bytes(
     return iter([_as_bytes(value) for value in values])
 
 
+def _set_as_bytes(
+    elements: TSetElementsInput, error_message: Optional[str] = DEFAULT_SET_CONVERSION_ERROR
+) -> TSetElementsInputBytes:
+    if not isinstance(elements, collections.abc.Iterable):
+        raise InvalidArgumentException(f"{error_message}{type(elements)}")
+    return iter({_as_bytes(element) for element in elements})
```
malandis, line 61: Why return `iter` as opposed to building a set?

src/momento/internal/aio/_scs_data_client.py
```diff
@@ -442,6 +457,88 @@ async def list_remove_value(
...
+        try:
+            self._log_issuing_request("SetAddElements", {})
+            _validate_cache_name(cache_name)
+            _validate_set_name(set_name)
+
+            item_ttl = self._default_ttl if ttl.ttl is None else ttl.ttl
+            request = _SetUnionRequest()
+            request.set_name = _as_bytes(set_name, self.__UNSUPPORTED_SET_NAME_TYPE_MSG)
+            request.elements.extend(_set_as_bytes(elements, self.__UNSUPPORTED_SET_ELEMENTS_TYPE_MSG))
+            request.ttl_milliseconds = int(item_ttl.total_seconds() * 1000)
```
malandis, line 476: We can refactor these using the method I introduced

src/momento/responses/set_data/set_fetch.py
```diff
@@ -0,0 +1,71 @@
...
+    See `SimpleCacheClient` for how to work with responses.
+    """
+
+
+class CacheSetFetch(ABC):
+    """Groups all `CacheSetFetchResponse` derived types under a common namespace."""
+
+    @dataclass
+    class Hit(CacheSetFetchResponse):
+        """Indicates the list exists and its values were fetched."""
```
malandis, line 28:
```suggestion
        """Indicates the set exists and its values were fetched."""
```

:scissors: :man_technologist: :notes:

src/momento/responses/set_data/set_fetch.py
```diff
@@ -0,0 +1,71 @@
...
+
+            Returns:
+                TSetElementsOutputStr
+            """
+
+            return {v.decode("utf-8") for v in self.elements}
+
+    @dataclass
+    class Miss(CacheSetFetchResponse):
+        """Indicates the list does not exist."""
```
malandis, line 58:
```suggestion
        """Indicates the set does not exist."""
```

src/momento/internal/synchronous/_scs_data_client.py
```diff
@@ -442,6 +457,88 @@ def list_remove_value(
...
+            request.set_name = _as_bytes(set_name, "Unsupported type for set_name: ")
+            response = self._build_stub().SetFetch(
+                request,
+                metadata=make_metadata(cache_name),
+                timeout=self._default_deadline_seconds,
+            )
+            self._log_received_response("SetFetch", {"set_name": str(request.set_name)})
+
+            type = response.WhichOneof("set")
+            if type == "missing":
```
cprice404, line 510: lol this still seems wild to me :)

src/momento/responses/set_data/set_fetch.py
```diff
@@ -0,0 +1,71 @@
...
+        """Indicates the list exists and its values were fetched."""
+
+        elements: Iterable[bytes]
+        """The elements as a iterable of bytes.
+
+        Use value_set_bytes and value_set_string to get the elements as a set.
+        """
+
+        @property
+        def value_set_bytes(self) -> TSetElementsOutputBytes:
```
cprice404, line 37: Is this type used in other files besides this one?  It's kind of a drag that I can't actually see the type definition in this file.  Especially in a PR, it means I have to jump to some other place to understand what's going on here.  Couldn't we just do the type aliases in the same files that consume them?

cprice404 (commented):
this looks good other than the fact that I don't love having to jump to a different file in order to see what the type alias evaluates to.  would like to hear your thoughts on making those more local so that the code is more readable.

also curious if you have looked to see what happens in an IDE like IntelliJ, whether the type hints that are shown are just showing the name of the alias, or the actual underlying type.  The latter would be far more valuable to users.

src/momento/simple_cache_client_async.py
```diff
@@ -601,11 +614,118 @@ async def list_remove_value(
...
+        ttl: CollectionTtl = CollectionTtl.from_cache_ttl(),
+    ) -> CacheSetAddElementResponse:
+        """
+        Adds an element to a set.
+
+        Args:
+            cache_name (TCacheName): The cache name with the set.
+            set_name (TSetName): The name of the set to add to.
+            element (TSetElement): The element to add.
+            ttl: (CollectionTtl): How to treat the list's TTL. Defaults to `CollectionTtl.from_cache_ttl()`
```
malandis, line 639:
```suggestion
            ttl: (CollectionTtl): How to treat the set's TTL. Defaults to `CollectionTtl.from_cache_ttl()`
```

src/momento/simple_cache_client_async.py
```diff
@@ -601,11 +614,118 @@ async def list_remove_value(
...
+        ttl: CollectionTtl = CollectionTtl.from_cache_ttl(),
+    ) -> CacheSetAddElementsResponse:
+        """
+        Add an elements to a set.
+
+        Args:
+            cache_name (TCacheName): The cache name with the set.
+            set_name (TSetName): The name of the set to add to.
+            elements (TSetElementsInput): The element to add.
+            ttl: (CollectionTtl): How to treat the list's TTL. Defaults to `CollectionTtl.from_cache_ttl()`
```
malandis, line 668:
```suggestion
            ttl: (CollectionTtl): How to treat the set's TTL. Defaults to `CollectionTtl.from_cache_ttl()`
```

tests/momento/simple_cache_client/test_set_async.py
```diff
@@ -0,0 +1,487 @@
...
+        assert isinstance(response, ErrorResponseMixin)
+        assert response.error_code == MomentoErrorCode.INVALID_ARGUMENT_ERROR
+        assert response.inner_exception.message == "Set name must be a string"
+
+
+@behaves_like(a_cache_name_validator)
+@behaves_like(a_connection_validator)
+@behaves_like(a_set_adder)
+@behaves_like(a_set_name_validator)
+def describe_set_add_element() -> None:
```
malandis, line 139: Also need to validate the element is a valid value

tests/momento/simple_cache_client/test_set_async.py
```diff
@@ -0,0 +1,487 @@
...
+        fetch_resp = await client_async.set_fetch(cache_name, set_name)
+        assert isinstance(fetch_resp, CacheSetFetch.Hit)
+        assert fetch_resp.value_set_bytes == {element1, element2}
+
+
+@behaves_like(a_cache_name_validator)
+@behaves_like(a_connection_validator)
+@behaves_like(a_set_adder)
+@behaves_like(a_set_name_validator)
+def describe_set_add_elements() -> None:
```
malandis, line 223: Need to validate `elements` is an iterable and contains valid values

tests/momento/simple_cache_client/test_set_async.py
```diff
@@ -0,0 +1,487 @@
...
+        client_async: SimpleCacheClientAsync, cache_name: TCacheName, set_name: TSetName
+    ) -> None:
+        resp = await client_async.set_fetch(cache_name, set_name)
+        assert isinstance(resp, CacheSetFetch.Miss)
+
+
+@behaves_like(a_cache_name_validator)
+@behaves_like(a_connection_validator)
+@behaves_like(a_set_name_validator)
+def describe_set_remove_element() -> None:
```
malandis, line 345: Validate element is valid

tests/momento/simple_cache_client/test_set_async.py
```diff
@@ -0,0 +1,487 @@
...
+
+        fetch_resp = await client_async.set_fetch(cache_name, set_name)
+        assert isinstance(fetch_resp, CacheSetFetch.Hit)
+        assert fetch_resp.value_set_bytes == new_elements
+
+
+@behaves_like(a_cache_name_validator)
+@behaves_like(a_connection_validator)
+@behaves_like(a_set_name_validator)
+def describe_set_remove_elements() -> None:
```
malandis, line 421: Validate elements are valid

malandis (commented):
Few changes

src/momento/responses/set_data/set_fetch.py
schwern: Works in VSCode.

Screen Shot 2023-02-03 at 15 59 38

src/momento/internal/_utilities/_data_validation.py
schwern: If grpc takes an iterator it will (presumably) stream instead of copying. Could potentially save a lot of memory.

src/momento/internal/aio/_scs_data_client.py
schwern: I put in #241 for that.

src/momento/internal/_utilities/_data_validation.py
malandis: Line 61 already builds a set object, so you're already paying that cost. You could change the function to be a generator if that's what you want:

```python
def set_elements_to_bytes(elements: List[bytes]) -> Iterator[bytes]:
    # check for collection type here
    for element in elements:
         yield _as_bytes(element)
```

src/momento/responses/set_data/set_fetch.py
```diff
@@ -0,0 +1,71 @@
...
+
+        @property
+        def value_set_string(self) -> TSetElementsOutputStr:
+            """The elements of the set, as utf-8 encoded strings.
+
+            Returns:
+                TSetElementsOutputStr
+            """
+
+            return {v.decode("utf-8") for v in self.elements}
```
malandis, line 54: The `elements` field should be named `value_set_bytes` and be of type `Set[bytes]`. Why have an iterator type field here when we're storing the set in memory anyway?

schwern: The FetchResponses should use iterators because it can get potentially huge, but I'm kinda half-assing it. I'll remove it for now.

src/momento/internal/_utilities/_data_validation.py
schwern: I do, but this iterator idea is going to take some more thought.

src/momento/responses/set_data/set_fetch.py
schwern: Oh, I remember. `elements` is just a raw dump of what's in the response and the response calls it `elements`. If the GRPC response is smart enough to stream the response, rather than load it all into memory, I'd like to take advantage of that. But, again, I'm half-assing this.

src/momento/responses/set_data/set_fetch.py
cprice404: it sounds like (from slack) that we tried this in PyCharm and it also works there?  If so I'm fine with it for now

src/momento/responses/set_data/set_fetch.py
cprice404: can we just make it a set for now?  We can always add some additional `value_iter` functions in the future if we have a request for it.

As a great man once said, YAGNI

cprice404 (commented):
i'm okay with this when @malandis is, I don't think we need to mess with `iter` for this go-around

src/momento/responses/set_data/set_fetch.py
schwern: elements is not a set, it's what we get back from the response.

malandis (approved):

cprice404 (approved):

PR #252, opened 2023-02-08
chore: undo public type aliases

malandis:
We use concrete type aliases instead of type aliases for the cache
client and responses. We do this because neither VSCode nor IntelliJ
resolve the type hints in methods when consuming the library. Ie when
typing `client.dictionary_fetch(` we do not see `cache_name: str` but
only `cache_name: TCacheName`.

### DIFF
--- a/src/momento/responses/list_data/list_fetch.py
+++ b/src/momento/responses/list_data/list_fetch.py
@@ -1,8 +1,9 @@
+from __future__ import annotations
+
 from abc import ABC
 from dataclasses import dataclass
 
 from momento.errors import SdkException
-from momento.typing import TListValuesOutputBytes, TListValuesOutputStr
 
 from ..mixins import ErrorResponseMixin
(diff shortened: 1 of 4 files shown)


schwern (changes requested):
There's still a few places where types are mentioned. I searched for `'T(Scalar|Cache|Dict|List|Collection)'` We agreed TDictionaryItems is ok.

* dictionary_fetch
* dictionary_get_fields
* list_fetch

And a comment in typing to remind anyone working on this that they are for internal use only, except TDictionaryItems.

malandis:
> There's still a few places where types are mentioned. I searched for `'T(Scalar|Cache|Dict|List|Collection)'` We agreed TDictionaryItems is ok.

Fixed in b7cbba5

cprice404 (approved):
thanks for doing this.  I'm good when @schwern is.

schwern (approved):

PR #257, opened 2023-02-08
refactor: Unroll the scalar ops.

### LINKED ISSUE #256: Extract the "common" method code.
The scalar ops are written in an outdated and complex fashion using "common" code split across several files and callbacks. That made sense when the async and sync clients had to be developed separately. Now that we're generating the sync client from the async we don't need it.

- [x] Eliminate the common scalar method code by just doing it in the scalar method.

schwern:
The common/ code was necessary to share between the syncronous and async clients. Now that we're generating the sync code from the async its a complexity we don't need. Unrolling the methods is much simpler.

Close #256


src/momento/internal/aio/_scs_data_client.py
```diff
@@ -141,13 +126,17 @@ class _ScsDataClient:
...
     def __init__(self, configuration: Configuration, credential_provider: CredentialProvider, default_ttl: timedelta):
         endpoint = credential_provider.cache_endpoint
         self._logger = logs.logger
         self._logger.debug("Simple cache data client instantiated with endpoint: %s", endpoint)
         self._endpoint = endpoint
 
-        default_deadline: timedelta = get_default_client_deadline(configuration)
+        default_deadline: timedelta = (
+            configuration.get_transport_strategy().get_grpc_configuration().get_deadline() or self._DEFAULT_DEADLINE
+        )
```
malandis, line 139: The deadline is not nullable in the config so this alternative shouldn't be necessary

malandis (commented):
One small change

malandis (approved):

PR #259, opened 2023-02-08
chore: add a hardcoded example for prepy310 to the README template

malandis:
Eventually we need to have multiple examples in the README:
- Python 3.10+ with pattern matching
- Python < 3.10 without pattern matching
- async client
- sync client

Since we are only displaying one example right now, this PR adds the sync
example for Python < 3.10. That way we have at least one example
compatible across all Python versions and one example with the pattern
matching.


cprice404 (approved):
thank you

PR #260, opened 2023-02-08
feat: Enforce keyword args for optional arguments.

### LINKED ISSUE #255: Use keywords for optional args in public methods.
For example, 

`def list_push_front(cache_name: str, list_name: str, value: str|bytes, *, ttl: CollectionTtl = CollectionTtl = CollectionTtl.from_cache_ttl(), truncate_back_to_size: Optional[int] = None)`

This forces the user to use keywords for the optional arguments making the order of arguments moot.

```
# fine
client.list_push_front(cname, lname, v)

# list_push_front() takes 3 positional arguments but 4 were given
client.list_push_front(cname, lname, v, ttl)

# fine
client.list_push_front(cname, lname, v, ttl=ttl)

# fine
client.list_push_front(cname, lname, v, truncate_back_to_size=ttl)

# fine
client.list_push_front(cname, lname, v, truncate_back_to_size=ttl, ttl=ttl)

# fine
client.list_push_front(cname, lname, v, ttl=ttl, truncate_back_to_size=ttl)
```

schwern:
This avoids having to worry about argument order.

I left set and set_if_not_exists alone becausae it would be a backwards incompat change.

Also
* set_add_elements() and set_remove_elements() was being passed a list, not a set.

Closes #255

### DIFF
--- a/src/momento/simple_cache_client.py
+++ b/src/momento/simple_cache_client.py
@@ -690,7 +690,7 @@ class SimpleCacheClient:
             CacheSetAddElementResponse
         """
 
-        resp = self.set_add_elements(cache_name, set_name, {element}, ttl=ttl)
+        resp = self.set_add_elements(cache_name, set_name, (element,), ttl=ttl)
 
         if isinstance(resp, CacheSetAddElements.Success):
             return CacheSetAddElement.Success()
@@ -750,7 +750,7 @@ class SimpleCacheClient:
         Returns:
             CacheSetRemoveElementResponse
         """
-        resp = self.set_remove_elements(cache_name, set_name, {element})
+        resp = self.set_remove_elements(cache_name, set_name, (element,))
         if isinstance(resp, CacheSetRemoveElements.Success):
             return CacheSetRemoveElement.Success()
         elif isinstance(resp, CacheSetRemoveElements.Error):
--- a/src/momento/simple_cache_client.py
+++ b/src/momento/simple_cache_client.py
@@ -180,7 +180,7 @@ class SimpleCacheClient:
         """
         return self._control_client.delete_cache(cache_name)
 
-    def list_caches(self, next_token: Optional[str] = None) -> ListCachesResponse:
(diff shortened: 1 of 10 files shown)


malandis:
> set_add_elements() and set_remove_elements() was being passed a list, not a set.

Why is this an issue?

schwern:
@malandis 

> > set_add_elements() and set_remove_elements() was being passed a list, not a set.
> 
> Why is this an issue?

Technically it isn't, but it gave me pause.

malandis:
> @malandis
> 
> > > set_add_elements() and set_remove_elements() was being passed a list, not a set.
> > 
> > 
> > Why is this an issue?
> 
> Technically it isn't, but it gave me pause.

If you want a really low overhead, trivial iterable, you can write: (element,). That is a 1-tuple which is iterable.

src/momento/simple_cache_client.py
```diff
@@ -682,7 +690,7 @@ def set_add_element(
             CacheSetAddElementResponse
         """
 
-        resp = self.set_add_elements(cache_name, set_name, [element], ttl)
+        resp = self.set_add_elements(cache_name, set_name, {element}, ttl=ttl)
```
malandis, line 693:
```suggestion
        resp = self.set_add_elements(cache_name, set_name, (element,), ttl=ttl)
```

src/momento/simple_cache_client.py
```diff
@@ -741,7 +750,7 @@ def set_remove_element(self, cache_name: str, set_name: str, element: str | byte
         Returns:
             CacheSetRemoveElementResponse
         """
-        resp = self.set_remove_elements(cache_name, set_name, [element])
+        resp = self.set_remove_elements(cache_name, set_name, {element})
```
malandis, line 753:
```suggestion
        resp = self.set_remove_elements(cache_name, set_name, (element,))
```

malandis (commented):
If you want a really low overhead, trivial iterable, you can write: `(element,)`. That is a 1-tuple which is iterable.

malandis (approved):

PR #262, opened 2023-02-09
refactor: Clean up the signatures of shared behaviors.

### LINKED ISSUE #254: Fix shared behavior type signatures.
- [ ] Switch to class based types which allow keyword arguments.
- [ ] Pass in consistent minimum arguments
  - [ ] Cache name
  - [ ] Collection name

schwern:
* Switch to a style that supports named arguments and is easier to understand.
* Only pass the necessary arguments to fixtures.
* For any test which needs to use the same cache, the fixture gets the cache name.

Also: other misc type issues in the tests.

Closes #254

### DIFF
--- a/tests/momento/simple_cache_client/shared_behaviors.py
+++ b/tests/momento/simple_cache_client/shared_behaviors.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from datetime import timedelta
 
 from typing_extensions import Protocol
@@ -7,8 +9,8 @@ from momento.auth import CredentialProvider
 from momento.config import Configuration
 from momento.errors import MomentoErrorCode
 from momento.responses import CacheResponse
-from momento.responses.mixins import ErrorResponseMixin
 from momento.typing import TCacheName, TScalarKey
+from tests.asserts import assert_response_is_error
 from tests.utils import uuid_str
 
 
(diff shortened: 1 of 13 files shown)


tests/momento/simple_cache_client/shared_behaviors.py
```diff
@@ -10,76 +11,97 @@
...
     def with_non_existent_cache_name_it_throws_not_found(cache_name_validator: TCacheNameValidator) -> None:
         cache_name = uuid_str()
-        response = cache_name_validator(cache_name)
-        assert isinstance(response, ErrorResponseMixin)
-        assert response.error_code == MomentoErrorCode.NOT_FOUND_ERROR
+        response = cache_name_validator(cache_name=cache_name)
+        if isinstance(response, ErrorResponseMixin):
+            assert response.error_code == MomentoErrorCode.NOT_FOUND_ERROR
+        else:
+            assert False
```
malandis, line 27: I don't understand this refactor (or the others that are like this)  :thinking:  Is there a typing issue, or something else? The previous version will give better detail when the assert fails and is also shorter

tests/momento/simple_cache_client/test_dictionary.py
```diff
@@ -287,7 +309,14 @@ def with_bytes_it_succeeds(
...
         dictionary_value_bytes: TDictionaryValue,
     ) -> None:
-        dictionary_setter(client, dictionary_name, dictionary_field_bytes, dictionary_value_bytes, ttl=CollectionTtl())
+        dictionary_setter(
+            client,
+            cache_name,
+            dictionary_name,
+            dictionary_field_bytes,
+            dictionary_value_bytes,
+            ttl=CollectionTtl(),
```
malandis, line 318: Can we omit the ttl from these (here and elsewhere) since the method gives them a default value?

tests/momento/simple_cache_client/test_list_async.py
```diff
@@ -186,48 +189,47 @@ async def with_other_values_type_it_errors(
         cache_name: TCacheName,
         list_name: TListName,
     ) -> None:
-        resp = await list_concatenator(cache_name, list_name, 234)
+        resp = await list_concatenator(cache_name, list_name, 234)  # type:ignore[arg-type]
```
malandis, line 192: :+1:

malandis (commented):
Nice improvement! Just a couple questions

tests/momento/simple_cache_client/shared_behaviors.py
schwern: Yes, without this `response` is a generic response and there's a type error. This uses type narrowing to solve that. There were so many type errors like this it was hard to work with on other type errors.

This happens so often a custom assert would be a better solution, but this fixes the type error.

tests/momento/simple_cache_client/test_dictionary.py
schwern: It's a setter and setters take a ttl, whether or not the tests currently use them. It's future proofing the interface.

schwern: Thinking about it, passing in a collectionTtl every time is probably a legacy of the previous way the types were set up. I think I can remove it.

tests/momento/simple_cache_client/shared_behaviors.py
malandis: I recommend just adding the assert back above the if statement and remove the else
```suggestion
        assert isinstance(response, ErrorResponseMixin)
        if isinstance(response, ErrorResponseMixin):
            assert response.error_code == MomentoErrorCode.NOT_FOUND_ERROR
```

tests/momento/simple_cache_client/test_dictionary.py
malandis: Can you do this for list and set as well?

tests/momento/simple_cache_client/shared_behaviors.py
schwern: I'd rather not have to repeat that sort of code everywhere. A little `def assert_response_is_error(response) -> ErrorResponseMixin:`  would solve the problem.

tests/momento/simple_cache_client/test_dictionary.py
schwern: I didn't see an issue in any of the other tests. The ttl has to remain in the signature of a collection setter, we just don't need to always pass it in.

tests/momento/simple_cache_client/shared_behaviors.py
schwern: I pushed a change which introduces a custom assert to handle this. I can't spend much time on it so I only applied it to the shared behaviors. The rest should be a SMOP.

malandis: TIL "SMOP" :smile:

malandis (approved):

PR #266, opened 2023-02-13
chore: enable linting on tests

### LINKED ISSUE #236: Fix type hints on tests and add tests folder to linting CI
- [x] Fix test type hints
- [x] Fix tests mypy errors
- [x] Enable mypy, flake8, black, isort on tests folder (also ensure `pyproject.toml` and `.flake8` are updated
- [x] Ensure CI also runs on tests folder consistent with Makefile

malandis:
This PR fixes the type hint errors throughout the tests and enables `mypy` and `flake8` on the tests in the Makefile and CI.

- chore: fix mypy errors on tests
- chore: enable mypy and flake8 on tests in Makefile
- chore: run mypy and flake8 on tests in CI

Closes #236

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,8 +1,10 @@
+from __future__ import annotations
+
 import asyncio
 import os
 import random
 from datetime import timedelta
-from typing import Optional, cast
+from typing import AsyncIterator, Callable, Iterator, List, Optional, Union, cast
 
 import pytest
 import pytest_asyncio
@@ -16,18 +18,12 @@ from momento.typing import (
     TDictionaryFields,
     TDictionaryItems,
     TDictionaryName,
-    TDictionaryValue,
     TListName,
     TListValue,
-    TListValuesInput,
-    TListValuesInputBytes,
-    TListValuesInputStr,
     TScalarKey,
     TScalarValue,
     TSetElement,
     TSetElementsInput,
-    TSetElementsInputBytes,
-    TSetElementsInputStr,
     TSetName,
 )
 from tests.utils import unique_test_cache_name, uuid_bytes, uuid_str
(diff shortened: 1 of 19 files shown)


tests/conftest.py
```diff
@@ -203,34 +203,37 @@ def event_loop() -> asyncio.AbstractEventLoop:  # type: ignore
 
 
 @pytest.fixture(scope="session")
-def client() -> SimpleCacheClient:
+def client() -> Iterator[SimpleCacheClient]:
```
cprice404, line 206: what's up with this iterator?

tests/conftest.py
```diff
@@ -203,34 +203,37 @@ def event_loop() -> asyncio.AbstractEventLoop:  # type: ignore
...
     configuration = Laptop.latest()
     credential_provider = CredentialProvider.from_environment_variable("TEST_AUTH_TOKEN")
     with SimpleCacheClient(configuration, credential_provider, DEFAULT_TTL_SECONDS) as _client:
         # Ensure test cache exists
-        _client.create_cache(TEST_CACHE_NAME)
+        _client.create_cache(cast(str, TEST_CACHE_NAME))
```
cprice404, line 211: why is this cast necessary?

tests/conftest.py
```diff
@@ -203,34 +203,37 @@ def event_loop() -> asyncio.AbstractEventLoop:  # type: ignore
...
-            await _client.delete_cache(TEST_CACHE_NAME)
+            await _client.delete_cache(cast(str, TEST_CACHE_NAME))
+
+
+TUniqueCacheName = Callable[[SimpleCacheClient], str]
 
 
 @pytest.fixture
-def unique_cache_name(client: SimpleCacheClient) -> str:
+def unique_cache_name(client: SimpleCacheClient) -> Iterator[Callable[[SimpleCacheClient], str]]:
```
cprice404, line 236: same question here about iterator

cprice404 (commented):
looks good!  just a few questions

tests/conftest.py
malandis: We read `TEST_CACHE_NAME` from the environment variable, which is an `Optional[str]` (ie nullable). Mypy complains unless I do this, even if I assert the variable is not `None`

tests/conftest.py
malandis: Same thing: because the fixture has setup and teardown, we use a yield statement (the setup goes before and teardown after). `Iterator` the preferred annotation for this; tests that consume the fixture only need to write `unique_cache_name: TUniqueCacheName` (see type alias just above this)

malandis: Same as the other fixture: because the fixture has setup and teardown, we use a yield statement (the setup goes before and teardown after). `Iterator` is the preferred annotation for this; tests that consume the fixture only need to write `unique_cache_name: TUniqueCacheName` (see type alias just above this)

malandis: For reference: https://docs.pytest.org/en/6.2.x/fixture.html

tests/conftest.py
cprice404: boo.  :)  wonder if we could assign it to a different variable that was explicitly annotated as a string after we handle the env var check.  not a blocker.

cprice404 (approved):

PR #272, opened 2023-02-17
chore: add client instantiation example to docs

nand4011:
Add an example of how to instantiate a client to the simple cache client docs.

Add a new string replacement to the code generation to replace any instance of 'SimpleCacheClientAsync' with 'SimpleCacheClient' to let the example doc be translated correctly.

### DIFF
--- a/src/momento/simple_cache_client_async.py
+++ b/src/momento/simple_cache_client_async.py
@@ -135,8 +135,8 @@ class SimpleCacheClientAsync:
                 It is possible to override this setting when calling the set method.
         Raises:
             IllegalArgumentException: If method arguments fail validations.
-        Example:
-        ::
+        Example::
+
             configuration = Laptop.latest()
             credential_provider = CredentialProvider.from_environment_variable("MOMENTO_AUTH_TOKEN")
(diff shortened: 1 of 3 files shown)


src/momento/simple_cache_client_async.py
```diff
@@ -135,6 +135,12 @@ def __init__(self, configuration: Configuration, credential_provider: Credential
                 It is possible to override this setting when calling the set method.
         Raises:
             IllegalArgumentException: If method arguments fail validations.
+        Example:
+        ::
```
malandis, line 139: Is the double colon a formatting thing?

nand4011: The double colon is the way I've seen recommended to display code examples in docstrings. I'm using it wrong here, however. It should be Example:: followed by a newline followed by the code. I wasn't using the newline and was getting weird formatting so I tried putting it on its own line and it worked. I'll alter it to be more correct.

malandis: Thank you! I was looking for this feature. Will update the rest of the docs accordingly.

malandis (approved):

PR #295, opened 2023-02-24
chore: improve imports for configurations and CredentialProvider

### LINKED ISSUE #284: improve the import situation for Configurations
Multiple devs reported issues trying to figure out how to import the configuration objects.

I think my preference would be to be able to do `from momento import configurations` and then in the constructor I would do `configurations.Laptop.v1()`, but if there is a more pythonic way to handle it that's fine.  At least needs to be more discoverable, perhaps by adding import statement to the docstring example code.

malandis:
Rolls up configurations and CredentialProvider to top level `momento`
package. Updates documentation of inline examples.

Closes #284

### DIFF
--- a/src/momento/__init__.py
+++ b/src/momento/__init__.py
@@ -1,7 +1,7 @@
 import logging
 
 from .auth import CredentialProvider
-from .config import configurations
+from .config import Configurations
 from .simple_cache_client import SimpleCacheClient
 from .simple_cache_client_async import SimpleCacheClientAsync
 
--- a/src/momento/__init__.py
+++ b/src/momento/__init__.py
@@ -1,5 +1,7 @@
 import logging
 
+from .auth import CredentialProvider
+from .config import configurations
 from .simple_cache_client import SimpleCacheClient
 from .simple_cache_client_async import SimpleCacheClientAsync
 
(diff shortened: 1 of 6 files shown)


src/momento/__init__.py
```diff
@@ -1,5 +1,7 @@
 import logging
 
+from .auth import CredentialProvider
+from .config import configurations
```
schwern, line 4: This is an oddball. Can it be Configurations?

schwern (commented):
This is so much better. I wonder if we can do something similar with responses?

src/momento/__init__.py
malandis: I hear you. We'd have to define an outer class in that module that contains all the pre-built configurations

malandis: Refactored in [afcef14](https://github.com/momentohq/client-sdk-python/pull/295/commits/afcef14a8c9787d05727dffb3131c47679c8fa78)

schwern (approved):

nand4011 (approved):
It's easy to miss the examples in the docs. Well caught.

malandis:
> This is so much better. I wonder if we can do something similar with responses?

I'm conflicted between making the responses easy to discover -- ie at the top level API -- vs grouping them nicely under cohesive package -- `momento.responses`

PR #302, opened 2023-02-27
chore: add trace level logging

nand4011:
Do nothing instead of raising an error if a log method is already defined in the add_logging_level method.

Call initialize_momento_logging in __init__.py to add trace logging.

Add a section to the readme that explains the trace logging and has an example of how to enable it.

### DIFF
--- a/README.template.md
+++ b/README.template.md
@@ -127,6 +127,17 @@ if __name__ == "__main__":
     _print_end_banner()
 ```
 
+### Logging
+To avoid cluttering DEBUG logging with per-method logs the Momento SDK adds a TRACE logging level. This will only happen
+if the TRACE level does not already exist.
+
+To enable TRACE level logging you can call logging.basicConfig() before making any log statements:
+```agsl
+import logging
+
+logging.basicConfig(level='TRACE')
+```
+
 ### Error Handling
 
 Coming Soon!
--- a/README.template.md
+++ b/README.template.md
@@ -134,7 +134,7 @@ if the TRACE level does not already exist.
 
 To enable TRACE level logging you can call logging.basicConfig() before making any log statements:
 
-```agsl
+```python
 import logging
 
 logging.basicConfig(level='TRACE')
--- a/README.template.md
+++ b/README.template.md
@@ -128,10 +128,12 @@ if __name__ == "__main__":
 ```
 
 ### Logging
+
 To avoid cluttering DEBUG logging with per-method logs the Momento SDK adds a TRACE logging level. This will only happen
 if the TRACE level does not already exist.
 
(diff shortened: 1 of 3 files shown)


src/momento/logs.py
```diff
@@ -38,12 +38,10 @@ def add_logging_level(level_name: str, level_num: int, method_name: Optional[str
     if not method_name:
         method_name = level_name.lower()
 
-    if hasattr(logging, level_name):
-        raise AttributeError("{} already defined in logging module".format(level_name))
-    if hasattr(logging, method_name):
-        raise AttributeError("{} already defined in logging module".format(method_name))
-    if hasattr(logging.getLoggerClass(), method_name):
-        raise AttributeError("{} already defined in logger class".format(method_name))
+    if hasattr(logging, level_name) or hasattr(logging, method_name) or hasattr(logging.getLoggerClass(), method_name):
```
nand4011, line 41: I suppose someone could make a trace logging level with a different method name than trace and we would be in trouble. Is there anything we can do about that?

README.template.md
```diff
@@ -127,6 +127,19 @@ if __name__ == "__main__":
...
 ```
 
+### Logging
+
+To avoid cluttering DEBUG logging with per-method logs the Momento SDK adds a TRACE logging level. This will only happen
+if the TRACE level does not already exist.
+
+To enable TRACE level logging you can call logging.basicConfig() before making any log statements:
+
+```agsl
```
malandis, line 137:
```suggestion
```python
```

agsl? :)

src/momento/logs.py
malandis: I think a warning would be appropriate. See [here](https://docs.python.org/3/library/warnings.html)

README.template.md
nand4011: Not sure how why the autocomplete assumed the example was in the Android Graphics Shading Language.

src/momento/logs.py
nand4011: Good idea

src/momento/logs.py
```diff
@@ -38,12 +39,15 @@ def add_logging_level(level_name: str, level_num: int, method_name: Optional[str
...
+        # Do nothing if the logging level already exists. Since this affects logging globally we can't know if the user
+        # or another dependency has already added the same level.
+        warnings.warn(
+            "Logging level {level} or logging method {method} already defined. Skipping.".format(
+                level=level_name, method=method_name
+            )
```
malandis, line 48: fyi can also interpolate the variables like so:

```python
warning.warn(f"Logging level {level} or ...")

malandis (approved):
✨

cprice404 (approved):

Issue #294, opened 2023-02-24
update examples post-manual-release


malandis:
- [x] new list caches api
- [x] configurations and CredentialProvider are in top level momento package
- [x] logging
- [x] v1 config
- [x] cache client rename
- [x] test names and types publicly visible in VS Code and IntelliJ
- [x] update inline example in README for all the above

https://github.com/momentohq/client-sdk-python/pull/279

PR #304, opened 2023-02-27
feat: rename cache client from `SimpleCacheClient` to `CacheClient`

### LINKED ISSUE #300: rename SimpleCacheClient to CacheClient and update examples

malandis:
- feat: rename `SimpleCacheClient` to `CacheClient`
- docs: update documentation to use `CacheClient`

We will update the examples post release. The issue tracking this is [here](https://github.com/momentohq/client-sdk-python/issues/294).

Closes #300

### DIFF
--- a/examples/prepy310/example.py
+++ b/examples/prepy310/example.py
@@ -3,7 +3,7 @@ from datetime import timedelta
 
 from example_utils.example_logging import initialize_logging
 
-from momento import SimpleCacheClient
+from momento import CacheClient
 from momento.auth import CredentialProvider
 from momento.config import Laptop
 from momento.responses import CacheGet, CacheSet, CreateCache, ListCaches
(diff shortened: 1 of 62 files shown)


examples/prepy310/example.py
```diff
@@ -3,7 +3,7 @@
 
 from example_utils.example_logging import initialize_logging
 
-from momento import SimpleCacheClient
+from momento import CacheClient
```
cprice404, line 6: you can't do the examples in the same PR, right?

malandis: Nope! My recursive `sed` call didn't know about that :P

malandis:
Note: I just renamed the cache client. We do not have a `SimpleCacheClient` alias.

cprice404 (approved):

PR #306, opened 2023-02-28
chore: add less verbose quickstart examples for README

cprice404:


malandis:
`make format` should fix this 👍

PR #310, opened 2023-02-28
docs: fix relative link to examples

malandis:


schwern:
The link Works For Me™ in main. Is this for compatibility with pydoc?

malandis:
> The link Works For Me™ in main. Is this for compatibility with pydoc?

The README is also [the description of the package on PyPi](https://pypi.org/project/momento/). Because of this, links have to be absolute to work there too.

schwern (approved):

PR #316, opened 2023-03-20
fix: bug where the __repr__ for Cred Provider would mutate it

cprice404:
Prior to this commit, the __repr__ function for a credentials
provider was grabbing a reference to the actual member variables
from the object before trying to obfuscate the token for printing.
Thus, any credential provider whose __repr__ was called would
no longer be valid for use with the cache client from that point
forward.

This commit ensures we do a deep copy of the attributes before
we obfuscate them, leaving the original attributes intact.
Also adds test coverage.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -198,7 +198,7 @@ Fear not! This is not twice the work. Since the synchronous client delegates beh
 to the asynchronous one, we can re-use the async test cases.
 
 When developing new test cases, only write tests for the async client in the appropriate
-`test_*_async.py` file. Then run `tests/scripts/sync_from_async.sh` to generate
-the synchronous client tests.
+`test_*_async.py` file. Then run `make do-gen-sync` to generate the synchronous client
+tests.
 
 <br/>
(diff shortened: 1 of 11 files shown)


malandis:
> This commit ensures we do a deep copy of the attributes before
we obfuscate them,

The change does a shallow copy; this is OK though since we just don't want to change the `vars` dictionary.

malandis (approved):

CONTRIBUTING.md
```diff
@@ -198,7 +198,7 @@ Fear not! This is not twice the work. Since the synchronous client delegates beh
 to the asynchronous one, we can re-use the async test cases.
 
 When developing new test cases, only write tests for the async client in the appropriate
-`test_*_async.py` file. Then run `tests/scripts/sync_from_async.sh` to generate
-the synchronous client tests.
+`test_*_async.py` file. Then run `make do-gen-sync` to generate the synchronous client
```
malandis, line 201: `do-gen-sync` is internal; `gen-sync` bundles this with running the formatter

```suggestion
`test_*_async.py` file. Then run `make gen-sync` to generate the synchronous client
```

pgautier404 (approved):

PR #319, opened 2023-04-12
feat: add support for v1 auth tokens

pgautier404:
This commit adds support for v1 auth tokens, similar to https://github.com/momentohq/client-sdk-nodejs/pull/336.

### DIFF
--- a/src/momento/auth/credential_provider.py
+++ b/src/momento/auth/credential_provider.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copy
 import os
 from dataclasses import dataclass
-from typing import Dict, Optional, Union
+from typing import Dict, Optional
 
 from . import momento_endpoint_resolver
 
--- a/src/momento/auth/credential_provider.py
+++ b/src/momento/auth/credential_provider.py
@@ -44,7 +44,7 @@ class CredentialProvider:
 
     @staticmethod
     def from_string(
-        auth_token: Union[bytes, str],
+        auth_token: str,
         control_endpoint: Optional[str] = None,
         cache_endpoint: Optional[str] = None,
     ) -> CredentialProvider:
--- a/src/momento/auth/credential_provider.py
+++ b/src/momento/auth/credential_provider.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copy
 import os
 from dataclasses import dataclass
-from typing import Dict, Optional
+from typing import Dict, Optional, Union
 
 from . import momento_endpoint_resolver
 
@@ -44,7 +44,7 @@ class CredentialProvider:
 
     @staticmethod
     def from_string(
-        auth_token: str,
+        auth_token: Union[bytes, str],
         control_endpoint: Optional[str] = None,
         cache_endpoint: Optional[str] = None,
     ) -> CredentialProvider:
@@ -60,9 +60,10 @@ class CredentialProvider:
         Returns:
             CredentialProvider
         """
-        endpoints = momento_endpoint_resolver.resolve(auth_token)
-        control_endpoint = control_endpoint or endpoints.control_endpoint
-        cache_endpoint = cache_endpoint or endpoints.cache_endpoint
+        token_and_endpoints = momento_endpoint_resolver.resolve(auth_token)
+        control_endpoint = control_endpoint or token_and_endpoints.control_endpoint
+        cache_endpoint = cache_endpoint or token_and_endpoints.cache_endpoint
+        auth_token = token_and_endpoints.auth_token
         return CredentialProvider(auth_token, control_endpoint, cache_endpoint)
 
     def __repr__(self) -> str:
--- a/src/momento/auth/momento_endpoint_resolver.py
+++ b/src/momento/auth/momento_endpoint_resolver.py
@@ -1,4 +1,7 @@
-from typing import Optional
+import base64
+import json
+from dataclasses import dataclass
+from typing import Union
 
 import jwt
 from jwt.exceptions import DecodeError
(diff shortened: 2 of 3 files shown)


src/momento/auth/momento_endpoint_resolver.py
```diff
@@ -11,27 +14,53 @@
...
         )
     except (DecodeError, KeyError) as e:
         raise InvalidArgumentException("Invalid Auth token.") from e
+
+
+def _is_base64(value: Union[bytes, str]) -> bool:
+    try:
+        if isinstance(value, str):
+            value = value.encode("utf-8")
+        return base64.b64encode(base64.b64decode(value)) == value
```
cprice404, line 64: this is a little janky but i realize you're just porting from matt's pr, and i missed it in that pr, and it doesn't affect the API, and it shouldn't get called often, so i guess carry on with this path for now and let's just get them done.

src/momento/auth/momento_endpoint_resolver.py
```diff
@@ -11,27 +14,53 @@
...
+    except (Exception,):
```
cprice404, line 65: what's up with the tuple here?

src/momento/auth/momento_endpoint_resolver.py
```diff
@@ -11,27 +14,53 @@
...
+    auth_token: str
 
 
-def resolve(auth_token: str, endpoint_override: Optional[str] = None) -> _Endpoints:
-    if endpoint_override is not None:
-        return _Endpoints(
-            _MOMENTO_CONTROL_ENDPOINT_PREFIX + endpoint_override,
-            _MOMENTO_CACHE_ENDPOINT_PREFIX + endpoint_override,
+@dataclass
+class Base64DecodedToken:
```
cprice404, line 25: i think you want v1 in this name in case there is a v2 later.  also @bruuuuuuuce do we have an easy way to differntiate those in the future?  presumably there is a version field we are just ignoring for now?

also should this class have an underscore prefix?

src/momento/auth/credential_provider.py
```diff
@@ -44,7 +44,7 @@ def from_environment_variable(
 
     @staticmethod
     def from_string(
-        auth_token: str,
+        auth_token: Union[bytes, str],
```
cprice404, line 47: why do we need to support bytes here?

tests/momento/auth/test_credential_provider.py
```diff
@@ -45,6 +53,44 @@
...
+        (
+            CredentialProvider.from_environment_variable(
+                env_var_name=test_v1_env_var_name,
+                control_endpoint="give.me.control.test.com",
+                cache_endpoint="secret.cache.test.com",
+            ),
+            test_v1_api_key,
+            "give.me.control.test.com",
+            "secret.cache.test.com",
+        ),
```
cprice404, line 93: does this cover all the test cases that @bruuuuuuuce added in his second PR against the node.js SDK?

cprice404:
would be good to test this by hand using a token you get from the console if it's easy.

cprice404 (commented):
handful of nits and questions but mostly looks fine!

src/momento/auth/momento_endpoint_resolver.py
pgautier404: I looked for a cleaner way to do this in python, and this seems to be a common approach. I didn't see anything that looked any better but am definitely open to suggestions and willing to do more looking.

src/momento/auth/momento_endpoint_resolver.py
pgautier404: This is just to suppress an IDE warning about `Exception` being overly broad. It's equivalent to `except Exception`.

src/momento/auth/momento_endpoint_resolver.py
pgautier404: Good call on both. Will rename.

src/momento/auth/credential_provider.py
pgautier404: This was also motivated by IDE warnings and was mostly intended to be defensive. After doing some testing though I'll go ahead and remove it as it's more complicated than needed.

tests/momento/auth/test_credential_provider.py
pgautier404: I believe so-- v1 vs legacy crossed with string vs env var crossed with endpoints overridden vs endpoints not overridden. @bruuuuuuuce please let me know if you see anything I'm missing here.

src/momento/auth/momento_endpoint_resolver.py
cprice404: not worth spending time on right now, just caught my eye.

tests/momento/auth/test_credential_provider.py
bruuuuuuuce: My only question is about this test 

```
CredentialProvider.from_string(auth_token=test_encoded_v1_token.decode("utf-8"))
```

Are we supporting decoding the tokens and passing in the decoded stringified json value to the credentials provider? If so I don't think we support this in nodejs yet

cprice404 (dismissed):
lgtm but get @bruuuuuuuce to sign off on test coverage :)

tests/momento/auth/test_credential_provider.py
pgautier404: Talked to Matt offline about this and we're all good. The conversion there is just for mypy's benefit.

bruuuuuuuce (approved):
:shipit:

PR #323, opened 2023-04-28
chore: Update the Momento wire types dependency

nand4011:
Update the Momento wire types dependency to allow the creation of sorted set methods.

Fix the increment request syntax that has changed with the new wire types version.

Avoid updating any other dependency because one of them causes a regression that prevents most of the tests from running.


malandis (approved):
Unfortunate about the regression. Looks like the rest of the SDKs will be affected by this as well.

PR #324, opened 2023-05-01
feat: Add sorted set put and fetch methods

nand4011:
Add sorted set put element and put elements.

Add sorted set fetch by rank and fetch by score.


src/momento/cache_client_async.py
```diff
@@ -770,6 +779,122 @@ async def set_remove_elements(
...
+
+    async def sorted_set_fetch_by_score(
+        self,
+        cache_name: str,
+        sorted_set_name: str,
+        min_score: float | None = None,
+        max_score: float | None = None,
+        sort_order: SortOrder = SortOrder.ASCENDING,
+        offset: int | None = None,
+        count: int | None = None,
```
malandis, line 846: The type hint for `type | None` is `Optional[type]`
```suggestion
        min_score: Optional[float] = None,
        max_score: Optional[float] = None,
        sort_order: SortOrder = SortOrder.ASCENDING,
        offset: Optional[int] = None,
        count: Optional[int] = None,
```

src/momento/cache_client_async.py
```diff
@@ -770,6 +779,122 @@ async def set_remove_elements(
...
+            min_score (float, optional): The minimum score of the range to fetch from the sorted set.
+                                         If None, fetches from the lowest score. Defaults to None.
+            max_score (float, optional): The maximum score of the range to fetch from the sorted set.
+                                         If None, fetches until the highest score. Defaults to None.
+            sort_order (SortOrder, optional): The sort order to use when fetching the sorted set.
+                                              Defaults to SortOrder.ASCENDING.
+            offset (int, optional): The number of elements to skip before starting to return elements.
+                                    If None, starts from the first element within the specified range. Defaults to None.
+            count (int, optional): The maximum number of elements to return.
+                                   If None, returns all elements within the specified range. Defaults to None.
```
malandis, line 862:
```suggestion
            min_score (Optional[float]): The minimum score of the range to fetch from the sorted set.
                                         If None, fetches from the lowest score. Defaults to None.
            max_score (Optional[float]): The maximum score of the range to fetch from the sorted set.
                                         If None, fetches until the highest score. Defaults to None.
            sort_order (SortOrder): The sort order to use when fetching the sorted set.
                                              Defaults to SortOrder.ASCENDING.
            offset (Optional[int]): The number of elements to skip before starting to return elements.
                                    If None, starts from the first element within the specified range. Defaults to None.
            count (Optional[int]): The maximum number of elements to return.
                                   If None, returns all elements within the specified range. Defaults to None.
```

src/momento/cache_client_async.py
```diff
@@ -770,6 +779,122 @@ async def set_remove_elements(
...
+        """Fetches a sorted set by rank range.
+
+        Args:
+            cache_name (str): Name of the cache containing the sorted set.
+            sorted_set_name (str): The name of the sorted set to fetch.
+            start_rank (int, optional): The start rank of the range to fetch from the sorted set.
+                                        If None, fetches from the beginning of the set. Defaults to None.
+            end_rank (int, optional): The end rank of the range to fetch from the sorted set.
+                                      If None, fetches until the end of the set. Defaults to None.
+            sort_order (SortOrder, optional): The sort order to use when fetching the sorted set.
```
malandis, line 888: Same suggestion as above.

src/momento/internal/aio/_scs_data_client.py
```diff
@@ -772,6 +793,160 @@ async def set_remove_elements(
...
+
+    async def sorted_set_fetch_by_score(
+        self,
+        cache_name: TCacheName,
+        sorted_set_name: TSortedSetName,
+        min_score: float | None,
+        max_score: float | None,
+        sort_order: SortOrder,
+        offset: int | None,
+        count: int | None,
```
malandis, line 840: Same suggestion regarding the Optional.

src/momento/internal/aio/_scs_data_client.py
```diff
@@ -772,6 +793,160 @@ async def set_remove_elements(
...
+            _validate_sorted_set_name(sorted_set_name)
+
+            request = _SortedSetFetchRequest()
+            request.set_name = _as_bytes(sorted_set_name, "Unsupported type for set_name: ")
+            request.with_scores = True
+
+            if min_score is not None:
+                request.by_score.min_score = min_score
+            else:
+                request.by_score.unbounded_min.CopyFrom(_Unbounded())
```
malandis, line 854: Is `CopyFrom` necessary or can we just directly assign `_Unbounded()` to `unbounded_min`

src/momento/internal/aio/_scs_data_client.py
```diff
@@ -772,6 +793,160 @@ async def set_remove_elements(
...
+
+            if max_score is not None:
+                request.by_score.max_score = max_score
+            else:
+                request.by_score.unbounded_max.CopyFrom(_Unbounded())
```
malandis, line 859: Same question about `unbounded_max`

src/momento/internal/aio/_scs_data_client.py
```diff
@@ -772,6 +793,160 @@ async def set_remove_elements(
...
+        except Exception as e:
+            self._log_request_error("sorted_set_fetch", e)
+            return CacheSortedSetFetch.Error(convert_error(e))
+
+    async def sorted_set_fetch_by_rank(
+        self,
+        cache_name: TCacheName,
+        sorted_set_name: TSortedSetName,
+        start_rank: int | None,
+        end_rank: int | None,
```
malandis, line 902: Same comment re: Optional

src/momento/internal/aio/_scs_data_client.py
```diff
@@ -772,6 +793,160 @@ async def set_remove_elements(
...
+
+            if start_rank is not None:
+                request.by_index.inclusive_start_index = start_rank
+            else:
+                request.by_index.unbounded_start.CopyFrom(_Unbounded())
+
+            if end_rank is not None:
+                request.by_index.exclusive_end_index = end_rank
+            else:
+                request.by_index.unbounded_end.CopyFrom(_Unbounded())
```
malandis, line 922: Same questions re: `CopyFrom`

src/momento/responses/data/sorted_set/fetch.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class CacheSortedSetFetch(ABC):
+    """Groups all `CacheSortedSetFetchResponse` derived types under a common namespace."""
+
+    @dataclass
+    class Hit(CacheSortedSetFetchResponse):
+        """Indicates the sorted set exists and its values were fetched."""
+
+        value_dictionary_bytes: dict[bytes, float]
```
malandis, line 29: We also want the array of value-score tuples.

tests/momento/simple_cache_client/test_sorted_set_async.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+class TSortedSetFetcher(Protocol):
+    def __call__(
+        self,
+        cache_name: TCacheName,
+        sorted_set_name: TSortedSetName,
+        sort_order: SortOrder,
+        minimum: Union[int, None] = None,
+        maximum: Union[int, None] = None,
```
malandis, line 227:
```suggestion
        minimum: Optional[int] = None,
        maximum: Optional[int] = None,
```

malandis (commented):
Mostly small changes but looks great.

src/momento/cache_client_async.py
malandis: The odd thing here in Python is that `Optional[float]` strictly means `float | None`, ie is nullable, and says nothing about whether it accepts a default value. Hence:

```python
def foo(x: Optional[int]) -> None
```

means `foo` is a void function that takes a single required argument `x`, which is a nullable int, vs

```python
def foo(x: Optional[int] = 42) -> None
```

means `foo` is a void function that takes an optional (ie not required) nullable int, which defaults to 42.

nand4011: I'll update the pr with optionals.

src/momento/responses/data/sorted_set/fetch.py
```diff
@@ -0,0 +1,68 @@
...
+            return [(key, value) for key, value in self.value_dictionary_bytes.items()]
+
+        @property
+        def value_list_string(self) -> list[tuple[str, float]]:
+            """The values for the fetched sorted set, as a list of utf-8 encoded string values to int score tuples.
+
+            Returns:
+                list[tuple[str, float]]
+            """
+            return [(key.decode("utf-8"), value) for key, value in self.value_dictionary_bytes.items()]
```
malandis, line 57: Since the order is lost when you store as a dictionary, could you: store the list instead of the dictionary as the response payload, then build the dictionaries on the fly? See [here](https://github.com/momentohq/client-sdk-python/blob/main/src/momento/responses/data/dictionary/get_fields.py) for an example where we want both the order preserved (list) and dictionary accessors.

malandis (approved):
❇️

PR #325, opened 2023-05-01
fix: remove manual release footgun

bruuuuuuuce:
The `on-push-to-release-branch` workflow should not be able to be run manually. It should only be run when a commit gets pushed to release


malandis (approved):
🖖

cprice404 (approved):

PR #327, opened 2023-05-03
feat: Add sorted set remove and increment

nand4011:
Add methods for removing elements from a sorted set.

Add a method incrementing a sorted set score.

### DIFF
--- a/src/momento/cache_client_async.py
+++ b/src/momento/cache_client_async.py
@@ -19,6 +19,7 @@ from momento.responses.data.sorted_set.get_scores import (
     CacheSortedSetGetScores,
     CacheSortedSetGetScoresResponse,
 )
+from momento.responses.data.sorted_set.increment import CacheSortedSetIncrementResponse
 from momento.responses.data.sorted_set.put_element import (
     CacheSortedSetPutElement,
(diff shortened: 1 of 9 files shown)


src/momento/cache_client_async.py
```diff
@@ -962,6 +971,72 @@ async def sorted_set_get_rank(
...
+        *,
+        ttl: CollectionTtl = CollectionTtl.from_cache_ttl(),
+    ) -> CacheSortedSetIncrementResponse:
+        """Increments the score for the given sorted set and value.
+
+        Args:
+            cache_name (str): Name of the cache to perform the lookup in.
+            sorted_set_name (str): The sorted set to look up.
+            value (str | bytes): The value in the sorted set to look up.
+            score (float): The quantity to add to the score. May be positive, negative, or zero. Defaults to 1.0.
```
malandis, line 1031: Can you also document what happens if a user increments the score of a value that doesn't exist?

malandis (commented):
Looks great. One small documentation request

src/momento/cache_client_async.py
nand4011: Will do.

malandis (approved):

PR #343, opened 2023-06-02
chore: restrict importlib_metadata package to python < 3.8

malandis:
Python 3.8 introduced a standard library `importlib.metadata` to
introspect package versions at runtime. Because of this, we only need
`importlib_metadata`, the backported library for python versions
<3.8. We modify the project requirements to only install this on those
versions, and to use the standard library when available.

### DIFF
--- a/src/momento/internal/_utilities/_momento_version.py
+++ b/src/momento/internal/_utilities/_momento_version.py
@@ -1,0 +1,14 @@
+momento_version = ""
+
+try:
+    # For python < 3.8
+    import importlib_metadata
+
+    momento_version = importlib_metadata.Distribution.from_name("momento").version  # type: ignore[no-untyped-call,misc]
+except ImportError:
+    # For python >= 3.8
+    from importlib.metdata import version  # type: ignore[import]
+
+    momento_version = version("momento")
+
+assert momento_version != ""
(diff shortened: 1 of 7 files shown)


src/momento/internal/_utilities/_momento_version.py
```diff
@@ -0,0 +1,14 @@
...
+    import importlib_metadata
+
+    momento_version = importlib_metadata.Distribution.from_name("momento").version  # type: ignore[no-untyped-call,misc]
+except ImportError:
+    # For python >= 3.8
+    from importlib.metadata import version  # type: ignore[import]
+
+    momento_version = version("momento")
+
+assert momento_version != ""
```
eaddingtonwhite, line 14: Hmm this assert makes me bit nervous maybe better fail open here default to something like "undefined" instead it's just for agent header right? If you feel good with hard fail here though I can get behind it.

eaddingtonwhite (dismissed):

src/momento/internal/_utilities/_momento_version.py
malandis: I will move this to a unit test

malandis: Done in [3a85dfb](https://github.com/momentohq/client-sdk-python/pull/343/commits/3a85dfb11ede1623439a27d6f8e4c174754cb2b9)

eaddingtonwhite (approved):

PR #354, opened 2023-07-13
chore: update examples for 1.7.0

pgautier404:


malandis (commented):
Should we add this to the workflows? eg https://github.com/momentohq/client-sdk-python/blob/main/.github/workflows/on-pull-request.yml#L141

eaddingtonwhite (approved):
@pgautier404 looks good. Only nit was trying to figure out where `cache` that this topic is on is created. Do we need a note to user to create ahead of time or code to create a cache to use? Fine w/ follow up pr for that though.

pgautier404:
> @pgautier404 looks good. Only nit was trying to figure out where `cache` that this topic is on is created. Do we need a note to user to create ahead of time or code to create a cache to use? Fine w/ follow up pr for that though.

Ugh, good catch, I'll go ahead and take care of that in this PR. Thanks.

pgautier404:
> Should we add this to the workflows? eg https://github.com/momentohq/client-sdk-python/blob/main/.github/workflows/on-pull-request.yml#L141

These examples are actually intended to be used in pairs, and are not suitable in their current form for CI use. The examples assume that a subscribe script in one terminal will receive and print messages from a publish script in a different terminal. None of the scripts both publish and subscribe. We can produce a CI-ready script to exercise the topic functionality end-to-end in a single go, but these are not intended for that use.

malandis (approved):
🥇

PR #358, opened 2023-08-15
feat: add vector index client

malandis:
Adds the new Vector Index Client, responses, and integration tests. Integration tests work against a local deployment of the service.

As such customers will not be able to use the client yet.


src/momento/internal/synchronous/_scs_grpc_manager.py
```diff
@@ -18,16 +18,24 @@
 from momento.internal.synchronous._retry_interceptor import RetryInterceptor
 from momento.retry import RetryStrategy
 
+# TODO: delete as this is for local testing only
+credentials = grpc.ssl_channel_credentials()
+try:
+    with open("./certs/server_cert.pem", "rb") as f:
```
malandis, line 24: Will revert this once we cut over to the live service.

src/momento/internal/aio/_scs_grpc_manager.py
```diff
@@ -19,6 +19,15 @@
...
 from ._retry_interceptor import RetryInterceptor
 
+# TODO: delete as this is for local testing only
+credentials = grpc.ssl_channel_credentials()
+try:
+    with open("./certs/server_cert.pem", "rb") as f:
+        cert = f.read()
+    credentials = grpc.ssl_channel_credentials(root_certificates=cert)
+except FileNotFoundError:
+    pass
```
malandis, line 29: Will revert this once we cut over to the live service.

src/momento/vector_index_client_async.py
```diff
@@ -0,0 +1,141 @@
...
+    raise e
+
+from momento.responses import (
+    CreateIndexResponse,
+    DeleteIndexResponse,
+    ListIndexesResponse,
+)
+
+
+class VectorIndexClientAsync:
```
malandis, line 41: 🎗️ rename to `PreviewVectorIndexClientAsync`

tests/conftest.py
```diff
@@ -38,22 +42,34 @@
...
 #######################
 
 TEST_CONFIGURATION = Configurations.Laptop.latest()
 TEST_TOPIC_CONFIGURATION = TopicConfigurations.Default.latest()
+TEST_VECTOR_CONFIGURATION = VectorIndexConfigurations.Default.latest()
 
 TEST_AUTH_PROVIDER = CredentialProvider.from_environment_variable("TEST_AUTH_TOKEN")
+TEST_VECTOR_AUTH_PROVIDER = CredentialProvider.from_environment_variable("TEST_AUTH_TOKEN")
+TEST_VECTOR_AUTH_PROVIDER.control_endpoint = "localhost:50051"
+TEST_VECTOR_AUTH_PROVIDER.cache_endpoint = "localhost:50052"
```
malandis, line 63: 🎗️ Will remove this once we cut over to the live service.

pgautier404 (dismissed):
LGTM, and I was able to run the tests against gRPC mock server.

src/momento/vector_index_client_async.py
```diff
@@ -0,0 +1,182 @@
...
+        print(
+            "There is an issue on M1 macs between GRPC native packaging and Python wheel tags. "
+            "See https://github.com/grpc/grpc/issues/28387",
+            file=sys.stderr,
+        )
+        print("-".join("" for _ in range(99)), file=sys.stderr)
+        print("    TO WORK AROUND:", file=sys.stderr)
+        print("    * Install Rosetta 2", file=sys.stderr)
+        print(
+            "    * Install Python from python.org (you might need to do this if you're using an arm-only build)",
```
cprice404, line 27: is this actually still a thing?

src/momento/cache_client.py
```diff
@@ -9,35 +9,6 @@
...
-    CacheSortedSetPutElementsResponse,
-)
-from momento.responses.data.sorted_set.remove_element import (
-    CacheSortedSetRemoveElement,
-    CacheSortedSetRemoveElementResponse,
-)
-from momento.responses.data.sorted_set.remove_elements import (
-    CacheSortedSetRemoveElements,
-    CacheSortedSetRemoveElementsResponse,
-)
```
cprice404, line 40: what's going on with these changes?

cprice404:
i kind of rushed through this, it's a big PR and I haven't had as much time to look at it as I need, but I didn't want to delay on getting some initial thoughts shared out

src/momento/internal/aio/_scs_control_client.py
```diff
@@ -26,6 +31,14 @@
...
     RevokeSigningKeyResponse,
 )
+from momento.responses.vector_index import (
+    CreateIndex,
+    CreateIndexResponse,
+    DeleteIndex,
+    DeleteIndexResponse,
+    ListIndexes,
+    ListIndexesResponse,
+)
```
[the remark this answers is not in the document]
malandis: Yes that's right

src/momento/config/transport/grpc_configuration.py
```diff
@@ -12,3 +13,11 @@ def get_deadline(self) -> timedelta:
...
     def with_deadline(self, deadline: timedelta) -> GrpcConfiguration:
         pass
+
+    @abstractmethod
+    def get_root_certificates(self) -> Optional[bytes]:
+        pass
+
+    @abstractmethod
+    def with_root_certificates(self, certificate: bytes) -> GrpcConfiguration:
+        pass
```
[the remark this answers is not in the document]
cprice404: i don't really love adding more surface area to the configuration object APIs without a customer-facing use case. also isn't your local setup going to become obsolete as soon as the real service goes live, since it's not the same tech stack?

src/momento/internal/aio/_scs_grpc_manager.py
```diff
@@ -28,7 +29,12 @@ class _ControlGrpcManager:
     def __init__(self, configuration: Configuration, credential_provider: CredentialProvider):
         self._secure_channel = grpc.aio.secure_channel(
             target=credential_provider.control_endpoint,
-            credentials=grpc.ssl_channel_credentials(),
+            credentials=grpc.ssl_channel_credentials(
+                # This was added to support local testing with a self-signed certificate.
+                root_certificates=configuration.get_transport_strategy()
+                .get_grpc_configuration()
+                .get_root_certificates()
+            ),
```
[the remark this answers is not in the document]
malandis: I intended to leave the grpc configuration stuff in to make local testing easier. I'm going to remove the hard coded provider info though.

src/momento/internal/aio/_vector_index_client.py
```diff
@@ -0,0 +1,105 @@
...
+from momento.responses.vector_index import (
+    Search,
+    SearchResponse,
+    UpsertItemBatch,
+    UpsertItemBatchResponse,
+)
+from momento.responses.vector_index.data.search import SearchHit
+
+
+class _VectorIndexClient:
```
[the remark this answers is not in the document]
malandis: I can do that. Here I was just paralleling the the name of the service in the protos.

src/momento/config/transport/grpc_configuration.py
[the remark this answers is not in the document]
cprice404: I won't block the PR on it if you feel strongly about it, it's not an unreasonable thing to support; I'm just doing my usual advocacy for keeping API footprint as small as possible

src/momento/cache_client.py
malandis: These hadn't been added to `momento.responses` for some reason, hence the really long `from momento.responses.data....`. While I was here I added those to `momento.responses` and moved these import lines below.

cprice404: :ack:

src/momento/config/vector_index_configurations.py
```diff
@@ -0,0 +1,33 @@
...
+        @staticmethod
+        def latest() -> VectorIndexConfigurations.Default:
+            """Provides the latest recommended configuration for a laptop development environment.
+
+            This configuration will be updated every time there is a new version of the laptop configuration.
+            """
+            return VectorIndexConfigurations.Default(
+                # This is high to account for time-intensive upserts.
+                StaticTransportStrategy(StaticGrpcConfiguration(timedelta(seconds=120))),
+                FixedCountRetryStrategy(max_attempts=3),
```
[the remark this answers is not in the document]
malandis: Thanks good catch!

src/momento/internal/aio/_vector_index_client.py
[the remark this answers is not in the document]
malandis: refactored in [a06c8c1](https://github.com/momentohq/client-sdk-python/pull/358/commits/a06c8c1f392d8575c395221262e5c3ee05759d74); agree for the sake of clarity it makes sense

src/momento/config/vector_index_configurations.py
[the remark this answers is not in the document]
malandis: Had to do some refactoring to accomodate this. Remember the cache control service hosts the vector index control RPCs. Thus to have a dedicated vector index configuration, I needed to create a separate vector index control client and control grpc manager. In [6bb890e](https://github.com/momentohq/client-sdk-python/pull/358/commits/6bb890ee6a718f8cfc8586f018930efdbbccdb94) I created modules to house each of these and refactored the code.

In [e82cd9f](https://github.com/momentohq/client-sdk-python/pull/358/commits/e82cd9fe7d096eb1f28d93c74fa36d1c919dfeca) I introduced the vector index configuration objects. These have a transport strategy and client timeout but not a retry strategy.

src/momento/config/transport/grpc_configuration.py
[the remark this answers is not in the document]
cprice404: @malandis did you have any more thoughts on this in response to comments above?

src/momento/config/vector_index_configuration.py
```diff
@@ -0,0 +1,63 @@
...
+    @abstractmethod
+    def with_transport_strategy(self, transport_strategy: TransportStrategy) -> VectorIndexConfigurationBase:
+        pass
+
+    @abstractmethod
+    def with_client_timeout(self, client_timeout: timedelta) -> VectorIndexConfigurationBase:
+        pass
+
+
+class VectorIndexConfiguration(VectorIndexConfigurationBase):
```
cprice404, line 23: thanks for doing this, I think this is a more future-proof path.

src/momento/internal/synchronous/_vector_index_data_client.py
```diff
@@ -0,0 +1,108 @@
...
+            self._build_stub().UpsertItemBatch(request, timeout=self._default_deadline_seconds)
+
+            self._log_received_response("UpsertItemBatch", {"index_name": index_name})
+            return UpsertItemBatch.Success()
+        except Exception as e:
+            self._log_request_error("set", e)
+            return UpsertItemBatch.Error(convert_error(e))
+
+    def search(
+        self, index_name: str, query_vector: list[float], top_k: int, metadata_fields: Optional[list[str]] = None
```
cprice404, line 67: I think it's probably okay to change this in a subsequent PR if we need to, but we might need a wrapper type for this `list[float]` here, right? for forward-compat with numpy etc.?

src/momento/vector_index_client_async.py
cprice404: I don't know if we need these rosetta workarounds anymore, I feel like this got fixed in gRPC. Can address in a subsequent PR but I would love to get rid of this.

cprice404 (approved):
i've left a few comments on here that I'm not sure we've bottomed out on. the biggest one is the gRPC transport changes involving the certs, because that is a public API whose scope extends beyond this new preview client, so if we add that we can't really change our mind about it.

nothing in my comments that I consider to be an absolute blocker but just make sure you feel really confident about that decision before we merge.

src/momento/vector_index_client_async.py
malandis: I don't know if it's still a thing. I copied it from the data client. Will be good to revisit and can do in a future PR.

src/momento/config/transport/transport_strategy.py
```diff
@@ -46,15 +47,25 @@ def with_client_timeout(self, client_timeout: timedelta) -> TransportStrategy:
 
 
 class StaticGrpcConfiguration(GrpcConfiguration):
-    def __init__(self, deadline: timedelta):
+    _deadline: timedelta
+    _root_certificates: Optional[bytes]
```
pratik151192, line 51: Would it make sense to have a separate gRPC configuration altogether for MVI as these don't seem related to cache? The only reason I'm saying is both these requirements may have divergent configurations over time, and I'm not sure you'd want to model all of them as `Optional` here because we don't want to break compatibility with the client configurations.

cprice404 (commented):
Do we have a handle on what the Red CI is about here?

I've chatted with Michael enough about this one to where I am basically ready to ship it once CI is green and he gives a 👍 .

My only real concerns are:
1. Distinct Configuration object for Vector rather than re-using Cache
2. New namespace for Vector responses, and a ticket created to do the same kind of namespacing (via aliases) for the existing response types
3. would prefer not to add new cert APIs to config for now if we don't need them, but it's not a blocker
4. naming conventions and other API considerations are reasonably aligned with Pete's JS PR. (there may be changes required in his PR to accommodate that, but I want to make sure they are aligned.)

malandis:
> Do we have a handle on what the Red CI is about here?
> 
> I've chatted with Michael enough about this one to where I am basically ready to ship it once CI is green and he gives a 👍 .
> 
> My only real concerns are:
> 
> 1. Distinct Configuration object for Vector rather than re-using Cache
We separate the config objects because the cache config had retries, and which aren't relevant to the vector client for now.  As for defaults, the vector add APIs have different latency requirements.

> 2. New namespace for Vector responses, and a ticket created to do the same kind of namespacing (via aliases) for the existing response types
> 3. would prefer not to add new cert APIs to config for now if we don't need them, but it's not a blocker
I thought we removed these? Let's double check for vestiges.

> 4. naming conventions and other API considerations are reasonably aligned with Pete's JS PR. (there may be changes required in his PR to accommodate that, but I want to make sure they are aligned.)
I left a common on his PR but we're otherwise aligned.

tests/utils.py
```diff
@@ -1,10 +1,18 @@
...
 
 
 def unique_test_cache_name() -> str:
     return f"python-test-{uuid_str()}"
 
 
+def unique_test_vector_index_name() -> str:
+    name = f"python-test-{uuid_str()}"
+    # TODO remove this
+    return name.replace("-", "_")
```
malandis, line 13:
```suggestion
    return f"python-test-{uuid_str()}"
```

We can remove that now :)

malandis (commented):
One small change to go

cprice404 (dismissed):

cprice404 (approved):

PR #365, opened 2023-08-21
feat: wire method to get and set eager connection timeout

pratik151192:
This PR adds the following to Python SDK

- Eager connection for both sync and async clients through a new factory method
- Lambda configuration
- Example for Python SDK and AWS Lambda
   - This will be updated to use the Lambda configuration created above in the next PR
  
### Testing

- Sync client:

``` poetry run pytest tests/momento/cache_client/test_scalar.py::describe_set_and_get_eager_connection_client --log-level=DEBUG```

```
------------------------------------------------------------ live log setup ------------------------------------------------------------
2023-08-22 11:06:51.829475 [DEBUG] Simple cache control client instantiated with endpoint: control.cell-4-us-west-2-1.prod.a.momentohq.com
2023-08-22 11:06:51.831961 [DEBUG] Simple cache data client instantiated with endpoint: cache.cell-4-us-west-2-1.prod.a.momentohq.com
2023-08-22 11:06:51.832132 [DEBUG] Attempting to create an eager connection with Momento's server within 10 seconds
2023-08-22 11:06:51.833295 [DEBUG] State is idle; waiting to transition to CONNECTING
2023-08-22 11:06:51.833431 [DEBUG] State transitioned to CONNECTING; waiting to get READY
2023-08-22 11:06:52.215337 [DEBUG] Connected to Momento's server! Happy Caching!
2023-08-22 11:06:52.215581 [INFO] Creating cache with name: default-cache

 tests/momento/cache_client/test_scalar.py::with_hit ✓                                                                    50% █████     

---------------------------------------------------------- live log teardown -----------------------------------------------------------
2023-08-22 11:06:53.408530 [INFO] Deleting cache with name: default-cache

2023-08-22 11:06:53.513750 [DEBUG] Closing and tearing down gRPC channel
 tests/momento/cache_client/test_scalar.py::with_byte_key_values ✓                                                       100% ██████████

Results (1.72s):
       2 passed
```

- Async client

```poetry run pytest tests/momento/cache_client/test_scalar_async.py::describe_set_and_get_eager_connection_client --log-level=DEBUG```

```
------------------------------------------------------------ live log setup ------------------------------------------------------------
2023-08-22 11:07:48.669870 [DEBUG] Using selector: KqueueSelector
2023-08-22 11:07:48.670081 [DEBUG] Using selector: KqueueSelector
2023-08-22 11:07:48.670324 [DEBUG] Simple cache control client instantiated with endpoint: control.cell-4-us-west-2-1.prod.a.momentohq.com
2023-08-22 11:07:48.670652 [DEBUG] Using AsyncIOEngine.POLLER as I/O engine
2023-08-22 11:07:48.671304 [DEBUG] Simple cache data client instantiated with endpoint: cache.cell-4-us-west-2-1.prod.a.momentohq.com
2023-08-22 11:07:48.671457 [DEBUG] Attempting to create an eager connection with Momento's server within 10 seconds
2023-08-22 11:07:48.671848 [DEBUG] State is idle; waiting to transition to CONNECTING
2023-08-22 11:07:48.672035 [DEBUG] State transitioned to CONNECTING; waiting to get READY
2023-08-22 11:07:49.025081 [DEBUG] Connected to Momento's server! Happy Caching!
2023-08-22 11:07:49.025377 [INFO] Creating cache with name: default-cache

 tests/momento/cache_client/test_scalar_async.py::with_hit ✓                                                              50% █████     

---------------------------------------------------------- live log teardown -----------------------------------------------------------
2023-08-22 11:07:50.298013 [INFO] Deleting cache with name: default-cache

2023-08-22 11:07:50.419840 [DEBUG] Closing and tearing down gRPC channel
2023-08-22 11:07:50.420181 [DEBUG] Using selector: KqueueSelector
2023-08-22 11:07:50.420344 [DEBUG] Using selector: KqueueSelector
 tests/momento/cache_client/test_scalar_async.py::with_byte_key_values ✓                                                 100% ██████████

Results (1.78s):
       2 passed

```

### CODE REVIEW

File: src/momento/internal/aio/_scs_grpc_manager.py [hunk 1]
```diff
@@ -64,8 +67,27 @@ def __init__(self, configuration: Configuration, credential_provider: Credential
                 # (experimental.ChannelOptions.SingleThreadedUnaryStream, 1)
             ],
         )
+        # we submit the eager connection task in our current running event loop
+        try:
+            self._eager_connect_task = asyncio.get_running_loop().create_task(self._eagerly_connect())
```
pratik151192, line 72: This was a bit interesting. The `asyncio` semantics doesn't provide a subscribe method similar to the sync gRPC channel. This part is non-blocking, and required some hacks (without changing the constructor to be async) to make it blocking per the timeout. So in essence this is best effort. Lmk if you think we should make the constructor async which will make it easier to block on this until a timeout.

src/momento/internal/synchronous/_scs_grpc_manager.py
```diff
@@ -46,15 +49,75 @@ class _DataGrpcManager:
...
+        EAGER_CONNECTION_TIMEOUT elapses.
+    """
+
+    def _eagerly_connect(self, configuration: Configuration) -> None:
+        eager_connection_timeout: timedelta = (
+            configuration.get_transport_strategy().get_grpc_configuration().get_eager_connection_timeout()
+        )
+
+        # An event to track whether we were able to establish an eager connection
+        connection_event = threading.Event()
```
pratik151192, line 76: We need this so that our `subscription` on line 115 doesn't lurk around even after it's job is done

examples/lambda/docker/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM --platform=linux/amd64 public.ecr.aws/lambda/python:3.8
+
+WORKDIR /var/task
+
+# Copy the lambda and the requirements file
+COPY lambda/index.py .
+COPY lambda/aws_requirements.txt .
+
+# Install Python dependencies
+RUN pip install -r aws_requirements.txt -t .
```
cprice404, line 10: @malandis this seems like a fine way to load deps for this lambda but if you have other thoughts let us know

pyproject.toml
```diff
@@ -59,6 +59,10 @@ libcst = "^0.4.9"
 
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
+log_level = "ERROR"
+log_cli = true
+log_cli_format = "%(asctime)s [%(levelname)s] %(message)s"
+log_cli_date_format = "%Y-%m-%d %H:%M:%S.%f"
```
cprice404, line 65: what is the scope of these settings? Is this for like test runs, or does this get included in the SDK and control logging config for users?

src/momento/internal/aio/_scs_grpc_manager.py
cprice404: I don't know if you even can make the constructor async, I don't know if the language allows it.

This is why, in the JS SDK, @nand4011 made a separate async factory function for instantiating the client. And that is also why the eager connection timeout did not end up being a new config setting at all, but rather, a new argument to that factory function.

My sense is that we need to follow that same path for the Python async API at least, and then for symmetry we would probably want to also add a factory function for the sync client.

I think this feature is kind of a non-starter if there is no way for the caller to block until the connection is established.

@malandis - thoughts?

cprice404 (commented):
just a few minor questions, with a little compare/contrast to JS

pyproject.toml
pratik151192: This is just for the test runs

src/momento/internal/aio/_scs_grpc_manager.py
pratik151192: > I don't know if you even can make the constructor async, I don't know if the language allows it.

I'll go over this once more but at a quick glance I saw some samples; might be interpreting it different though.

> I think this feature is kind of a non-starter if there is no way for the caller to block until the connection is established.

I honestly had the same instinct and hence mentioned best-effort but now since you also called it out, it's not that tricky to make it blocking with the factory pattern you mentioned.

> then for symmetry we would probably want to also add a factory function for the sync client.

For the sync client, the reason I didn't think of that pattern as de-facto was because in NodeJS we **had** to follow that pattern as we couldn't do it in the existing constructor. In DotNet, I believe we enable this feature for free. And I believe we wanted to do the same in the SDKs where it was doable, so that customers get eager connections for free.

cprice404: I actually think that if I could go back in time I might have done the .NET one differently. We're making a network call so it seems pretty reasonable for it to surface to the end user as an async thing that they get to decide how they want to handle.

But @malandis is our most experienced python dev so it would be good to get his input.

malandis: Agreed. Since the init methods are not async, we should not be awaiting anything in them, even if in a roundabout way. A factory method is a better here.

pyproject.toml
pratik151192: To verify you can look at line 60 `tool.pytest.ini_options`

examples/lambda/README.md
```diff
@@ -0,0 +1,45 @@
...
+<img src="https://docs.momentohq.com/img/logo.svg" alt="logo" width="400"/>
+
+[![project status](https://momentohq.github.io/standards-and-practices/badges/project-status-official.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+[![project stability](https://momentohq.github.io/standards-and-practices/badges/project-stability-stable.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+
+<br>
+
+## Example Lambda
+
+This repo contains an example lambda, built using AWS CDK, that performs a basic set and get operation on Momento cache.
```
cprice404, line 13:
```suggestion
This directory contains an example lambda, built using AWS CDK, that performs a basic set and get operation on Momento cache.
```

examples/lambda/docker/lambda/index.py
```diff
@@ -0,0 +1,36 @@
...
+from momento.responses import CacheGet, CacheSet, CreateCache
+
+
+def handler(event, lambda_context):
+    cache_name = "default-cache"
+    with CacheClient(
+            configuration=Configurations.Laptop.v1(),
+            credential_provider=CredentialProvider.from_environment_variable("MOMENTO_AUTH_TOKEN"),
+            default_ttl=timedelta(seconds=60),
+    ) as cache_client:
```
cprice404, line 13: so we will change this to use your new stuff once we have a release?

pratik151192: correct

examples/lambda/infrastructure/lib/stack.ts
```diff
@@ -0,0 +1,21 @@
...
+export class MomentoLambdaStack extends cdk.Stack {
+    constructor(scope: Construct, id: string, props?: cdk.StackProps) {
+        super(scope, id, props);
+
+        // Create Lambda function from Docker Image
+        const dockerLambda = new lambda.DockerImageFunction(this, 'MomentoDockerLambda', {
+            functionName: 'MomentoDockerLambda',
+            code: lambda.DockerImageCode.fromImageAsset(path.join(__dirname, '../../docker')), // Point to the root since Dockerfile should be there
+            environment: {
+                MOMENTO_AUTH_TOKEN: process.env.MOMENTO_AUTH_TOKEN || ''
```
cprice404, line 15: should you throw if this env var is not set? The lambda is not going to work so you might save someone some heartache by failing to deploy it?

pyproject.toml
cprice404: 🤦 duh thank you

cprice404 (commented):
this lgtm. i added a couple of wording suggestion commits. would be good to get @malandis to give a quick scan for API considerations, and wouldn't hurt for @nand4011 to give a quick scan too, but if they aren't available by tomorrow I think we can probably move forward.

eager

nand4011 (commented):
LGTM. I can approve after Chris' comments are resolved.

examples/lambda/docker/Dockerfile
malandis: LGTM

pyproject.toml
malandis: Very cool!

src/momento/cache_client_async.py
```diff
@@ -186,6 +190,45 @@ def __init__(self, configuration: Configuration, credential_provider: Credential
...
+
+            configuration = Configurations.Laptop.latest()
+            credential_provider = CredentialProvider.from_environment_variable("MOMENTO_AUTH_TOKEN")
+            ttl_seconds = timedelta(seconds=60)
+            eager_connection_timeout = timedelta(seconds=30)
+            client = CacheClientAsync.create(configuration, credential_provider, ttl_seconds, eager_connection_timeout)
+        """
+        validate_eager_connection_timeout(eager_connection_timeout)
+        # an explicit 0 means that the client disabled eager connections
+        if eager_connection_timeout.seconds != 0:
```
malandis, line 226: This is unlikely to surface as a bug, given most timeouts will be small. This should be:

```suggestion
        if eager_connection_timeout.total_seconds() != 0:
```

`.seconds` represents the seconds portion of a structured timedelta object as opposed to the total number of seconds expressed by the time duration. See [here](https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds), [here](https://docs.python.org/3/library/datetime.html#timedelta-objects), and .NET's `TimeSpan` for a similar gotcha.

Python seems to fold minutes and hours to seconds, but leaves days as is. So `abs(timeout.seconds) < 60 seconds * 60 minutes * 24 hours`.

This is a common source of bugs in customer code and one to keep in mind!

src/momento/internal/aio/_scs_grpc_manager.py
```diff
@@ -65,7 +68,41 @@ def __init__(self, configuration: Configuration, credential_provider: Credential
             ],
         )
 
+    async def eagerly_connect(self, timeout_seconds: int) -> None:
+        self._logger.debug(
+            f"Attempting to create an eager connection with Momento's server within {timeout_seconds} seconds"
+        )
+        try:
+            await asyncio.wait_for(self.wait_for_ready(), timeout_seconds)
```
malandis, line 76: Very nice!

src/momento/utilities/shared_sync_asyncio.py
```diff
@@ -0,0 +1,8 @@
+from datetime import timedelta
+
+DEFAULT_EAGER_CONNECTION_TIMEOUT_SECONDS = 30
+
+
+def validate_eager_connection_timeout(timeout: timedelta) -> None:
+    if timeout.total_seconds() < 0:
```
malandis, line 7: 👍

src/momento/cache_client_async.py
```diff
@@ -186,6 +190,45 @@ def __init__(self, configuration: Configuration, credential_provider: Credential
             for _ in range(CacheClientAsync._NUM_CLIENTS)
         ]
 
+    @staticmethod
+    async def create(
```
malandis, line 194: Is this how we've called this in other libraries?

malandis (commented):
I found a minor bug but otherwise looks good!

examples/lambda/infrastructure/lib/stack.ts
pratik151192: That's a great suggestion. Will do!

src/momento/cache_client_async.py
pratik151192: Thanks Michael! Very informative!

src/momento/cache_client_async.py
pratik151192: Yes!

malandis (approved):

PR #373, opened 2023-08-28
chore: use vector endpoint for MVI API calls

pratik151192:
For MVI data plane calls, we were using `cache` as the endpoint prefix; this PR updates it to use `vector`. Note that for legacy tokens, we do not have a new JWT claim, so we will be using `cache` as the prefix for that as well.

Also, for control plane operations, the endpoint prefix is still `control`.

### DIFF
--- a/src/momento/auth/momento_endpoint_resolver.py
+++ b/src/momento/auth/momento_endpoint_resolver.py
@@ -11,14 +11,17 @@ from momento.internal.services import Service
 
 _MOMENTO_CONTROL_ENDPOINT_PREFIX = "control."
 _MOMENTO_CACHE_ENDPOINT_PREFIX = "cache."
+_MOMENTO_VECTOR_ENDPOINT_PREFIX = "vector."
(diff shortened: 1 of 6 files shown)


src/momento/auth/momento_endpoint_resolver.py
```diff
@@ -11,14 +11,17 @@
 
 _MOMENTO_CONTROL_ENDPOINT_PREFIX = "control."
 _MOMENTO_CACHE_ENDPOINT_PREFIX = "cache."
+_MOMENTO_VECTOR_ENDPOINT_PREFIX = "vector."
 _CONTROL_ENDPOINT_CLAIM_ID = "cp"
 _CACHE_ENDPOINT_CLAIM_ID = "c"
+_VECTOR_ENDPOINT_CLAIM_ID = "c"
```
pratik151192, line 17: we dont have a new claim here so defaulting to `c`

malandis (approved):
LGTM. @cprice404 ?

cprice404 (approved):
alright

PR #385, opened 2023-10-05
chore: prefer API key over AUTH token terminology

malandis:
We have settled on the terminology "API_KEY" instead of
"AUTH_TOKEN". This updates the examples, documentation, and workflows
to use that. An exception are the tests which are internal.

Over the course of this refactor, linting and formatting failed for the examples, so I have corrected errors as well.

### DIFF
--- a/.github/workflows/on-pull-request-mvi.yml
+++ b/.github/workflows/on-pull-request-mvi.yml
@@ -61,7 +61,7 @@ jobs:
         run: /Users/runner/.local/bin/poetry run black src tests --check --diff
 
       - name: Run isort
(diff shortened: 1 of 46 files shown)


nand4011 (approved):
Looks like credential_provider still uses "auth token" in docstrings.

nand4011 (approved):

malandis:
> Looks like credential_provider still uses "auth token" in docstrings.

Done. Will have to leave the argument name as `auth_token` for backwards compatibility.

anitarua (approved):

cprice404 (commented):
thank you!

PR #390, opened 2023-10-10
chore: add doc examples for the vector index client

nand4011:


examples/py310/doc-examples-python-apis.py
```diff
@@ -166,6 +169,84 @@ async def example_API_TopicPublish(topic_client: TopicClientAsync):
...
+        VectorIndexConfigurations.Default.latest(), CredentialProvider.from_environment_variable("MOMENTO_API_KEY")
+    )
+
+
+# end example
+
+async def example_API_CreateIndex(vector_client: PreviewVectorIndexClientAsync):
+    response = await vector_client.create_index("test-index", 2)
+    match response:
+        case CreateIndex.Success:
```
malandis, line 183: Python's match statements are kind of odd. Each of the case statements needs to be an instance of the object:
```suggestion
        case CreateIndex.Success():
```

examples/py310/doc-examples-python-apis.py
```diff
@@ -166,6 +169,84 @@ async def example_API_TopicPublish(topic_client: TopicClientAsync):
...
+            print("Index 'test-index' created")
+        case CreateIndex.IndexAlreadyExists:
+            print("Index 'test-index' already exists")
+        case CreateIndex.Error:
+            print("Error creating index 'test-index': ", response.message)
```
malandis, line 188: For the error statements, could you use the convention elsewhere in the file, eg:

```suggestion
        case CreateIndex.Error() as error:
            print(f"Error creating index 'test-index': {error.message}")
```

malandis (commented):
Looks good just a couple changes!

examples/py310/doc-examples-python-apis.py
nand4011: Good catch.

examples/py310/doc-examples-python-apis.py
nand4011: That is much nicer.

examples/py310/doc-examples-python-apis.py
```diff
@@ -160,12 +163,90 @@ async def example_API_TopicPublish(topic_client: TopicClientAsync):
     match response:
         case TopicPublish.Success():
             print("Successfully published a message")
-        case TopicPublish.Error():
-            print("Error publishing a message: ", response.message)
+        case TopicPublish.Error() as error:
+            print("Error publishing a message: ", error.message)
```
malandis, line 167: For consistency, can you use f strings for the error print statements? 🐍

malandis (commented):
Just one style point about f strings

examples/py310/doc-examples-python-apis.py
nand4011: Done!

malandis (approved):

PR #392, opened 2023-10-20
feat: support more vector item metadata value types

malandis:
In the previous version of the service, we only supported
string-valued metadata. In the latest version, we support the
following value types:

- string
- int
- float
- bool
- list of strings

We update the `Item` class to support serializing these types and the
`SearchHit` class to deserialize.

### DIFF
--- a/src/momento/requests/vector_index/item.py
+++ b/src/momento/requests/vector_index/item.py
@@ -19,7 +19,7 @@ class Item:
     vector: list[float]
     """The vector of the item."""
 
-    metadata: dict[str, str] = field(default_factory=dict)
+    metadata: dict[str, str | int | float | bool | list[str]] = field(default_factory=dict)
     """The metadata of the item."""
 
     def to_proto(self) -> pb._Item:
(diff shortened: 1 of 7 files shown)


src/momento/requests/vector_index/item.py
```diff
@@ -19,19 +19,29 @@ class Item:
     vector: list[float]
     """The vector of the item."""
 
-    metadata: dict[str, str] = field(default_factory=dict)
+    metadata: dict[str, str | int | float | bool | list[str]] = field(default_factory=dict)
```
malandis, line 22: I left the types enumerated like this because the IDE (VSCode, PyCharm) will not expand type aliases in an overlay. I can extract it to a type alias if you feel strongly about it.

src/momento/responses/vector_index/data/search.py
```diff
@@ -24,15 +26,27 @@ class SearchResponse(VectorIndexResponse):
 class SearchHit:
     id: str
     distance: float
-    metadata: dict[str, str] = field(default_factory=dict)
+    metadata: dict[str, str | int | float | bool | list[str]] = field(default_factory=dict)
```
malandis, line 29: Same comment here re: type alias.

src/momento/requests/vector_index/item.py
```diff
@@ -19,19 +19,29 @@ class Item:
...
     """The metadata of the item."""
 
     def to_proto(self) -> pb._Item:
         vector = pb._Vector(elements=self.vector)
         metadata = []
         if self.metadata is not None:
             for k, v in self.metadata.items():
-                if type(v) is not str:
+                if type(v) == str:
```
pratik151192, line 30: nit: Do you think it will be cleaner if we had a `ser_de_utility` file that handles the serialization/deserialization and this path just call that? Usually when I am reading code in a new repo, it helps seeing both the methods in a single file even if different files (or classes) are accessing them.

pratik151192: Another style nit that usually I have seen `switch/case` being used in such cases for readability. But i dont have a strong opinion on that.

tests/momento/requests/vector_index/test_item.py
```diff
@@ -5,22 +5,45 @@
...
+        id="id",
+        vector=[1, 2, 3],
+        metadata={
+            "string_key": "value",
+            "int_key": 1,
+            "double_key": 3.14,
+            "bool_key": True,
+            "list_of_strings_key": ["one", "two"],
+        },
+    ).to_proto()
```
pratik151192, line 19: Just for my knowledge - is there an advantge in converting back to `proto` to test fields?

eaddingtonwhite (approved):
🚢 Nice job on tests! And wrastling w/ the grpc `oneof` types 😓

pratik151192 (approved):

tests/momento/requests/vector_index/test_item.py
malandis: I wanted to specifically test we were converting to protobuf correctly, as I've had typos/copypasta errors in the past with oneofs.

src/momento/requests/vector_index/item.py
malandis: re serialization utility: maybe? While it's a cross cutting concern and we could centralize it, my first instinct was to place it with the class for cohesion. Can you show me what you mean tomorrow?

re switch/case: this syntax (`match`) wasn't introduced until Python 3.10. Because we have to code to lowest common denominator so that our one library works on older and newer versions, we have to write if-else statements. Unfrotunately.

src/momento/responses/vector_index/data/search.py
cprice404: if it's just 2 places then i agree with the choice to bias towards IDE compat

cprice404 (approved):

PR #393, opened 2023-10-23
chore: explicitly test behavior when using default similarity metric

malandis:
Previously we verified that we can create an index with the default
similarity metric. In this commit we add a test to explicitly check
that the behavior is consistent with `COSINE_SIMILARITY`.

### DIFF
--- a/tests/momento/vector_index_client/test_data.py
+++ b/tests/momento/vector_index_client/test_data.py
@@ -1,5 +1,9 @@
 from __future__ import annotations
 
+from typing import Optional
+
+import pytest
+
 from momento import PreviewVectorIndexClient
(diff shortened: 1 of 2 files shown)


pratik151192:
Accidentally closed the issue. 

For my knowledge, what was the issue when rishti tried to create the index ?

pratik151192 (approved):

malandis:
> For my knowledge, what was the issue when rishti tried to create the index ?

Still unclear

pratik151192:
@malandis 

> Still unclear

We discussed this in sprint planning today. My bad - that was the JS SDK actually that Rishti was using through console and I think Chris is looking into that.

PR #394, opened 2023-10-23
feat: allow custom root certificates in vector index client config

malandis:
Adds a `with_root_certificates_pem` to read a certificate pem
file. Exposes a `with_root_certificates_pem` method to the top level
of `VectorIndexConfiguration` for convenience. The clients are
instantiated with either the default ssl channel creds or ones using
the root certificates, if available.

### DIFF
--- a/src/momento/config/transport/grpc_configuration.py
+++ b/src/momento/config/transport/grpc_configuration.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from datetime import timedelta
+from pathlib import Path
+from typing import Optional
 
 
 class GrpcConfiguration(ABC):
@@ -11,4 +13,12 @@ class GrpcConfiguration(ABC):
 
     @abstractmethod
     def with_deadline(self, deadline: timedelta) -> GrpcConfiguration:
+        pass
+
+    @abstractmethod
+    def with_root_certificates_pem(self, root_certificates_pem_path: Path) -> GrpcConfiguration:
+        pass
+
+    @abstractmethod
+    def get_root_certificates_pem(self) -> Optional[bytes]:
         pass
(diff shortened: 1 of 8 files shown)


src/momento/config/transport/grpc_configuration.py
```diff
@@ -12,3 +14,11 @@ def get_deadline(self) -> timedelta:
...
     def with_deadline(self, deadline: timedelta) -> GrpcConfiguration:
         pass
+
+    @abstractmethod
+    def with_root_certificates_pem(self, root_certificates_pem_path: Path) -> GrpcConfiguration:
+        pass
+
+    @abstractmethod
+    def get_root_certificates_pem(self) -> Optional[bytes]:
+        pass
```
malandis, line 24: I added these to the `GrpcConfiguration` class and implemented in the `StaticGrpcConfiguration` class, which is used across cache, topic, and vector clients. I've only plumbed through using the root certificates in the vector clients though.

As these are buried deep, I feel that most users won't use it; and if one does and needs the functionality in other clients, we can plumb it through there. Otherwise I will either (1) plumb through in other clients now, or (2) specialize a `GrpcConfiguration` subtype that contains the the data and only use that in the vector client.

cprice404: no this is fine

src/momento/internal/aio/_vector_index_grpc_manager.py
```diff
@@ -19,7 +22,7 @@ class _VectorIndexControlGrpcManager:
     def __init__(self, configuration: VectorIndexConfiguration, credential_provider: CredentialProvider):
         self._secure_channel = grpc.aio.secure_channel(
             target=credential_provider.control_endpoint,
-            credentials=grpc.ssl_channel_credentials(),
+            credentials=vector_credentials_from_root_certs_or_default(configuration),
```
cprice404, line 25: so this is what you were saying about this will only work with vector for now?

i'd be fine with you making that util function general and using in other clients but it's not a blokcer for this pr

cprice404 (dismissed):

src/momento/internal/aio/_vector_index_grpc_manager.py
malandis: Yes that's right. Can open up in a follow up PR.

eaddingtonwhite (approved):

PR #397, opened 2023-11-04
feat: add mvi search score threshold

malandis:
Add the ability to specify a score threshold to prune search results
server side. Search hit distances must meet this threshold to be
included.

For euclidean similarity, scores lower than the threshold are
included; for inner product and cosine similarity, higher scores are
included.

Bumps the protos, implements the feature, and adds tests.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -280,13 +280,13 @@ files = [
 
 [[package]]
 name = "momento-wire-types"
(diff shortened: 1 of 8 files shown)


malandis:
@nand4011 will need this in the .NET SDK too.

nand4011 (approved):

PR #399, opened 2023-11-08
feat: add mvi search and fetch vectors method

malandis:
Add MVI `search_and_fetch_vectors` method and integration tests. This new method performs a `search` as usual, but returns the vectors associated with the search hits. That is, the only change in behavior comes down to the response. 

Because of this, we refactor the search tests to work against both `search` and `search_and_fetch_vectors`. We make use of pytest's parameterized test feature to do this.

### DIFF
--- a/src/momento/internal/aio/_vector_index_data_client.py
+++ b/src/momento/internal/aio/_vector_index_data_client.py
@@ -19,12 +19,12 @@ from momento.responses.vector_index import (
     DeleteItemBatchResponse,
     Search,
     SearchAndFetchVectors,
-    SearchResponse,
+    SearchAndFetchVectorsHit,
     SearchAndFetchVectorsResponse,
+    SearchHit,
+    SearchResponse,
     UpsertItemBatch,
     UpsertItemBatchResponse,
-    SearchHit,
-    SearchAndFetchVectorsHit,
 )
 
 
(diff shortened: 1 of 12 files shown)


src/momento/internal/aio/_vector_index_data_client.py
```diff
@@ -136,6 +139,53 @@ async def search(
...
+            else:
+                metadata_fields_pb = vectorindex_pb._MetadataRequest(
+                    some=vectorindex_pb._MetadataRequest.Some(
+                        fields=metadata_fields if metadata_fields is not None else []
+                    )
+                )
+
+            no_score_threshold = None
+            if score_threshold is None:
+                no_score_threshold = vectorindex_pb._NoScoreThreshold()
```
pratik151192, line 167: Not 100% on what significance `no_score_threshold` has here, why initialized as `None` and then asigned an empty `vectorindex_pb._NoScoreThreshold()` if the `score_threshold` is `None`

nand4011: It looks like either score_threshold or no_score_threshold must be set in the request object. If score_threshold is None, no_score_threshold is set, otherwise it is left as None. Then both are added to the request object.

nand4011 (approved):

src/momento/internal/aio/_vector_index_data_client.py
malandis: Protobuf in python forces you to initialize `oneof` fields in the message constructor. In our case there either is a score threshold or there isn't.

If there is a score threshold, then we need to set `score_threshold` to that value and `no_score_threshold` to `None` (indicating it is not part of the oneof).

If there isn't a score threshold, then we set `score_threshold` to `None` and we instantiate `no_score_threshold`.

pratik151192 (approved):

PR #410, opened 2023-11-21
chore: use ruff linter

malandis:
[Ruff](https://github.com/astral-sh/ruff) is a drop in replacement for flake8, isort, black, and their various plugins. This simplifies the toolchain, is much faster to run, and does much more comprehensive static analysis.

In this PR we've added ruff and corrected errors.


bruuuuuuuce (approved):
:shipit:

PR #411, opened 2023-11-21
chore: use ruff formatter and remove black

malandis:
Use the built-in ruff formatter and remove the black formatter from
the project. With this change only mypy and ruff are needed for (1)
type checking, and (2) static analysis and formatting, respectively.


bruuuuuuuce (approved):
:shipit:

PR #414, opened 2023-11-21
ci: fix example clobbering

malandis:
The examples use a cache called `test-cache`. This is hard-coded and
easy to read for a nice user experience. Because of this, we need to
run the examples sequentially or else one job may delete the cache
while another job is still working.

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -78,6 +78,10 @@ jobs:
             package: py310
           - python-version: "3.11"
             package: py310
+      # For simplicity, the examples use a cache called "test-cache"
+      # Therefore we cannot run them in parallel or else we could prematurely delete
+      # the cache for one job in another job.
+      max-parallel: 1
 
     env:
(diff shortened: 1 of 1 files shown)


bruuuuuuuce:
couldnt it still get clobbered if 2 or more prs are opened and running the examples at the same time?

malandis:
@bruuuuuuuce That's a good point. I'd prefer not to complicate the examples by reading the cache name from the environment; I can't think of another way to solve that problem otherwise, can you?

bruuuuuuuce:
> @bruuuuuuuce That's a good point. I'd prefer not to complicate the examples by reading the cache name from the environment; I can't think of another way to solve that problem otherwise, can you?

Why not have them just generate a random guid?

malandis:
@bruuuuuuuce GUID isn't very user friendly for a getting started, and we don't use a GUID in JS examples (for comparison). Let's do this for now and can loop back if we concurrent PRs are an issue.

bruuuuuuuce (approved):
:shipit:

PR #416, opened 2023-11-28
chore: bump mypy dependency

malandis:
Mypy has been 1.0 and VS Code plugins require 1.0. This PR bumps the
mypy dep.


bruuuuuuuce (approved):
:shipit:

PR #417, opened 2023-11-29
feat: implement get item batch and variant

malandis:
Adds MVI methods for get item batch and get item and fetch vectors
batch and integration tests.

### DIFF
--- a/src/momento/responses/vector_index/data/item.py
+++ b/src/momento/responses/vector_index/data/item.py
@@ -1,0 +1,28 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+
+@dataclass
+class Item:
+    """Represents an entry in the vector index."""
+
+    id: str
+    """The id of the item."""
+
+    vector: list[float]
+    """The vector of the item."""
+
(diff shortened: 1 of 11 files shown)


src/momento/responses/vector_index/data/item.py
```diff
@@ -0,0 +1,28 @@
...
+
+@dataclass
+class ItemWithoutVector:
+    """Represents an entry in the vector index without a vector."""
+
+    id: str
+    """The id of the item."""
+
+    metadata: dict[str, str | int | float | bool | list[str]] = field(default_factory=dict)
+    """The metadata of the item."""
```
malandis, line 28: In a follow-up PR I will refactor `SearchHit`, `SearchAndFetchVectorsHit`, `UpsertItemBatch` to use these as well.

nand4011 (approved):
LGTM, other than the monstrous mixing of camel case and snake case in the log messages ;-)

pgautier404 (approved):

malandis:
> LGTM, other than the monstrous mixing of camel case and snake case in the log messages ;-)

Good eye. One set is for the RPCs, which are PascalCase; the other set are for the python methods, which are snake_case.

PR #418, opened 2023-11-30
refactor: explicitly model item vs item metadata

malandis:
A common concern between requests and responses is a full `Item` in
the index vs `Metadata` vs `ItemMetadata` (id + metadata). We refactor
the code to model this. This is a backwards compatible change.


pgautier404 (approved):
LGTM!

PR #430, opened 2024-02-06
feat: mvi id in set filter expression

malandis:
Bumps the python proto dependency and implements the MVI filter expression `IdInSet`. This filter expression is satisfied when an item has an id that is in the set provided by the user. We also add integration tests for this filter.


eaddingtonwhite (approved):
🧼

PR #435, opened 2024-02-12
refactor: rename mvi parameter `ids` to `filter` on `get`/`delete` items

malandis:
- refactor: rename `ids` parameter to `filter` on delete_item_batch
- refactor: rename `ids` parameter to `filter` on get item batch
- refactor: rename `ids parameter to `filter` on `get item metadata`

### DIFF
--- a/examples/prepy310/vector_index.py
+++ b/examples/prepy310/vector_index.py
@@ -107,7 +107,7 @@ def search(client: PreviewVectorIndexClient, index_name: str) -> None:
 def delete_items(client: PreviewVectorIndexClient, index_name: str) -> None:
     item_ids_to_delete = ["test_item_1", "test_item_3"]
     _logger.info(f"Deleting items: {item_ids_to_delete}")
-    delete_response = client.delete_item_batch(index_name, ids=item_ids_to_delete)
+    delete_response = client.delete_item_batch(index_name, filter=item_ids_to_delete)
     if isinstance(delete_response, DeleteItemBatch.Success):
         _logger.info("Successfully deleted items")
     elif isinstance(delete_response, DeleteItemBatch.Error):
--- a/src/momento/vector_index_client.py
+++ b/src/momento/vector_index_client.py
@@ -277,17 +277,17 @@ class PreviewVectorIndexClient:
             index_name, query_vector, top_k, metadata_fields, score_threshold, filter
         )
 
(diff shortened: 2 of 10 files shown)


examples/prepy310/vector_index.py
```diff
@@ -107,7 +107,7 @@ def search(client: PreviewVectorIndexClient, index_name: str) -> None:
 def delete_items(client: PreviewVectorIndexClient, index_name: str) -> None:
     item_ids_to_delete = ["test_item_1", "test_item_3"]
     _logger.info(f"Deleting items: {item_ids_to_delete}")
-    delete_response = client.delete_item_batch(index_name, ids=item_ids_to_delete)
+    delete_response = client.delete_item_batch(index_name, filter=item_ids_to_delete)
```
bruuuuuuuce, line 110: just wondering, why `filter` and not `filters`? Looks like this is an array

src/momento/vector_index_client.py
```diff
@@ -194,19 +194,19 @@ def upsert_item_batch(self, index_name: str, items: list[Item]) -> UpsertItemBat
...
-    def delete_item_batch(self, index_name: str, ids: list[str]) -> DeleteItemBatchResponse:
+    def delete_item_batch(self, index_name: str, filter: list[str]) -> DeleteItemBatchResponse:
         """Deletes a batch of items from a vector index.
 
         Deletes any and all items with the given IDs from the index.
 
         Args:
             index_name (str): Name of the index to delete the items from.
-            ids (list[str]): The IDs of the items to be deleted from the index.
+            filter (list[str]): The IDs of the items to be deleted from the index.
```
bruuuuuuuce, line 204: should we rename these arg descriptions to 

```
filter (list[str]): The filters used to delete items from the index
```

examples/prepy310/vector_index.py
malandis: In a coming PR this will accept a (single) filter expression or (for convenience) a list of ids to select.

src/momento/vector_index_client.py
malandis: In the future state this will be a filter expression. We will keep accepting a list of id's because that is a common case so we want to make it easy without wrapping them in a filter object.

bruuuuuuuce (approved):
:shipit:

PR #441, opened 2024-02-26
feat: add grpc config options and turn off keepalive for Lambda config

anitarua:
Adds keepalive and max message size grpc channel options. Turns off keepalive for the Lambda cache configuration.

### DIFF
--- a/src/momento/config/configurations.py
+++ b/src/momento/config/configurations.py
@@ -35,14 +35,7 @@ class Configurations:
             This configuration is guaranteed not to change in future releases of the Momento Python SDK.
             """
             return Configurations.Laptop(
-                StaticTransportStrategy(
-                    StaticGrpcConfiguration(
-                        deadline=timedelta(seconds=15),
-                        keepalive_permit_without_calls=True,
-                        keepalive_time=timedelta(milliseconds=5000),
-                        keepalive_timeout=timedelta(milliseconds=1000),
-                    )
-                ),
+                StaticTransportStrategy(StaticGrpcConfiguration(deadline=timedelta(seconds=15))),
                 FixedCountRetryStrategy(max_attempts=3),
             )
 
@@ -60,6 +53,8 @@ class Configurations:
                     StaticGrpcConfiguration(
                         deadline=timedelta(milliseconds=1100),
                         keepalive_permit_without_calls=False,
+                        keepalive_time=None,
+                        keepalive_timeout=None,
                     )
                 ),
                 FixedCountRetryStrategy(max_attempts=3),
@@ -93,14 +88,7 @@ class Configurations:
                 This configuration is guaranteed not to change in future releases of the Momento Python SDK.
                 """
                 return Configurations.InRegion.Default(
-                    StaticTransportStrategy(
-                        StaticGrpcConfiguration(
-                            deadline=timedelta(milliseconds=1100),
-                            keepalive_permit_without_calls=True,
-                            keepalive_time=timedelta(milliseconds=5000),
-                            keepalive_timeout=timedelta(milliseconds=1000),
-                        )
-                    ),
+                    StaticTransportStrategy(StaticGrpcConfiguration(deadline=timedelta(milliseconds=1100))),
                     FixedCountRetryStrategy(max_attempts=3),
                 )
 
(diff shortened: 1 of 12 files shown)


cprice404 (dismissed):
this one LGTM! but i'd love @malandis to take a peek when he has time.

alive

src/momento/config/configurations.py
```diff
@@ -35,7 +35,14 @@ def v1() -> Configurations.Laptop:
             This configuration is guaranteed not to change in future releases of the Momento Python SDK.
             """
             return Configurations.Laptop(
-                StaticTransportStrategy(StaticGrpcConfiguration(timedelta(seconds=15))),
+                StaticTransportStrategy(
+                    StaticGrpcConfiguration(
+                        deadline=timedelta(seconds=15),
+                        keepalive_permit_without_calls=True,
+                        keepalive_time=timedelta(milliseconds=5000),
+                        keepalive_timeout=timedelta(milliseconds=1000),
```
malandis, line 43: It looks like these three settings are identical to every instance except one. I recommend making these the default values in the constructor then and we can specialize in just the instance we make for the lambda config.

src/momento/config/transport/transport_strategy.py
```diff
@@ -71,6 +85,21 @@ def with_root_certificates_pem(self, root_certificates_pem_path: Path) -> GrpcCo
     def get_root_certificates_pem(self) -> Optional[bytes]:
         return self._root_certificates_pem
 
+    def get_max_send_message_length(self) -> Optional[int]:
+        return self._max_send_message_length if self._max_send_message_length is not None else None
+
+    def get_max_receive_message_length(self) -> Optional[int]:
+        return self._max_receive_message_length if self._max_receive_message_length is not None else None
```
malandis, line 92: Can just return the attribute here since the attribute is either an int or `None`.

```suggestion
    def get_max_send_message_length(self) -> Optional[int]:
        return self._max_send_message_length

    def get_max_receive_message_length(self) -> Optional[int]:
        return self._max_receive_message_length
```

src/momento/config/transport/grpc_configuration.py
```diff
@@ -22,3 +22,23 @@ def with_root_certificates_pem(self, root_certificates_pem_path: Path) -> GrpcCo
...
+    def get_keepalive_permit_without_calls(self) -> Optional[int]:
+        pass
+
+    @abstractmethod
+    def get_keepalive_time(self) -> Optional[int]:
+        pass
+
+    @abstractmethod
+    def get_keepalive_timeout(self) -> Optional[int]:
+        pass
```
malandis, line 44: These should return a `Optional[timedelta]`, right?

src/momento/config/transport/transport_strategy.py
```diff
@@ -71,6 +85,21 @@ def with_root_certificates_pem(self, root_certificates_pem_path: Path) -> GrpcCo
...
+
+    def get_keepalive_permit_without_calls(self) -> Optional[int]:
+        return int(self._keepalive_permit_without_calls) if self._keepalive_permit_without_calls is not None else None
+
+    def get_keepalive_time(self) -> Optional[int]:
+        return self._keepalive_time.seconds * 1000 if self._keepalive_time is not None else None
+
+    def get_keepalive_timeout(self) -> Optional[int]:
+        return self._keepalive_timeout.seconds * 1000 if self._keepalive_timeout is not None else None
```
malandis, line 101: I would expect these to return an `Optional[timedelta]`. Do you want methods that return the time in milliseconds?

malandis (commented):
Few questions about the new methods

src/momento/internal/_utilities/_grpc_channel_options.py
```diff
@@ -0,0 +1,46 @@
...
+    if keepalive_permit is not None:
+        channel_options.append(("grpc.keepalive_permit_without_calls", keepalive_permit))
+
+    keepalive_time = grpc_config.get_keepalive_time()
+    if keepalive_time is not None:
+        channel_options.append(("grpc.keepalive_time_ms", keepalive_time.seconds * 1000))
+
+    keepalive_timeout = grpc_config.get_keepalive_timeout()
+    if keepalive_timeout is not None:
+        channel_options.append(("grpc.keepalive_timeout_ms", keepalive_timeout.seconds * 1000))
```
malandis, line 44: On a closer look, this conversion from `timedelta (seconds)` to milliseconds is not correct. `timedelta.seconds` returns the total seconds up until a day and does not account for milliseconds:

- `timedelta(days=1, minutes=1).seconds` is 60
- `timedelta(milliseconds=500).seconds` is 0
- `timedelta(milliseconds=1100)`.seconds is 1

This has implications in the rest of the code (cache client in particular). I can address this in a followup PR.

malandis (approved):

PR #442, opened 2024-03-04
fix: timedelta to ms conversion and timedelta to deadline conversion

malandis:
This PR corrects two bugs.

First we add a utility function to convert timedelta objects to milliseconds, unit tests, and refactor code to use this across the board. Previously we were converting timedeltas to milliseconds in several places which led to bugs when calling `.seconds` (not fractional) vs `total_seconds()` (fractional). Calling the former was truncating the milliseconds portion.

Second we correct a bug the gRPC timeout. After reading from the config, we converted the timedelta to total whole seconds. This again truncated the milliseconds portion; because the gRPC parameter accepts fractional seconds, we should not have been truncating.

### DIFF
--- a/src/momento/internal/aio/_scs_data_client.py
+++ b/src/momento/internal/aio/_scs_data_client.py
@@ -16,6 +16,7 @@ from momento.internal._utilities import (
     _gen_dictionary_items_as_bytes,
     _gen_list_as_bytes,
     _gen_set_input_as_bytes,
+    _timedelta_to_ms,
     _validate_cache_name,
     _validate_dictionary_name,
     _validate_list_name,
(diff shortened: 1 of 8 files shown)


src/momento/internal/aio/_scs_data_client.py
```diff
@@ -154,7 +155,7 @@ def __init__(self, configuration: Configuration, credential_provider: Credential
         self._endpoint = endpoint
 
         default_deadline: timedelta = configuration.get_transport_strategy().get_grpc_configuration().get_deadline()
-        self._default_deadline_seconds = int(default_deadline.total_seconds())
+        self._default_deadline_seconds = default_deadline.total_seconds()
```
malandis, line 158: grpc's timeout takes fractional seconds to account for milliseconds, and `timedelta.total_seconds` returns fractional seconds, so we needn't truncate

anitarua (approved):
good catch!!

PR #443, opened 2024-03-04
fix: disable keepalive in grpc channel options for control plane clients and set grpc channel options in synchronous code

anitarua:
Follow-up to #441 after we determined control plane connections should not have keepalives.

Also propagating the grpc channel options changes over to the synchronous code as well.

### DIFF
--- a/src/momento/internal/_utilities/_grpc_channel_options.py
+++ b/src/momento/internal/_utilities/_grpc_channel_options.py
@@ -8,11 +8,14 @@ from momento.internal._utilities import _timedelta_to_ms
 DEFAULT_MAX_MESSAGE_SIZE = 5_243_000  # bytes
 
 
-def grpc_channel_options_from_grpc_config(grpc_config: GrpcConfiguration) -> grpc.aio.ChannelArgumentType:
+def grpc_channel_options_from_grpc_config(
+    grpc_config: GrpcConfiguration, is_control_client: bool = False
+) -> grpc.aio.ChannelArgumentType:
     """Create gRPC channel options from a GrpcConfiguration.
 
     Args:
         grpc_config (GrpcConfiguration): the gRPC configuration.
+        is_control_client (bool, optional): whether the client is a control client, in which case we want to disable keepalives. Defaults to False.
 
     Returns:
         grpc.aio.ChannelArgumentType: a list of gRPC channel options as key-value tuples.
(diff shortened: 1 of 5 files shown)


src/momento/internal/aio/_scs_grpc_manager.py
```diff
@@ -34,13 +34,23 @@ class _ControlGrpcManager:
     version = momento_version
 
     def __init__(self, configuration: Configuration, credential_provider: CredentialProvider):
+        # Copy over all grpc configs from the vector config, but disable keepalive settings
```
malandis, line 37: Typo? This is the cache control client.

malandis (commented):
Just one question. Also run `make gen-sync` to generate the synchronous client code from the async code.

src/momento/internal/_utilities/_grpc_channel_options.py
```diff
@@ -8,11 +8,14 @@
 DEFAULT_MAX_MESSAGE_SIZE = 5_243_000  # bytes
 
 
-def grpc_channel_options_from_grpc_config(grpc_config: GrpcConfiguration) -> grpc.aio.ChannelArgumentType:
+def grpc_channel_options_from_grpc_config(
+    grpc_config: GrpcConfiguration, is_control_client: bool = False
+) -> grpc.aio.ChannelArgumentType:
```
malandis, line 13: Is it not an issue that this is scoped under the `aio` package but we're using it for both the async and sync clients?

src/momento/internal/_utilities/_grpc_channel_options.py
```diff
@@ -8,11 +8,14 @@
...
     """Create gRPC channel options from a GrpcConfiguration.
 
     Args:
         grpc_config (GrpcConfiguration): the gRPC configuration.
+        is_control_client (bool, optional): whether the client is a control client, in which case we want to disable keepalives. Defaults to False.
```
malandis, line 18: nit: for clarity's sake I prefer separate functions, eg:

`grpc_data_channel_options_from_grpc_config`

`grpc_control_channel_options_from_grpc_config`

And it's ok for the implementation of the latter to just invoke the former and override settings where appropriate.

malandis (commented):
1 question and 1 style nit otherwise looks good

src/momento/internal/_utilities/_grpc_channel_options.py
anitarua: I'm not sure, it seems like it's just a type alias but and `grpc.secure_channel` in the sync code takes `Any | None` for options, but yeah idk how it resolves. Will do some local testing

> `(type alias) ChannelArgumentType: type[Sequence[Tuple[str, Any]]]`

src/momento/internal/_utilities/_grpc_channel_options.py
```diff
@@ -1,14 +1,17 @@
...
-import grpc
+from typing import Sequence, Tuple, Union
 
 from momento.config.transport.grpc_configuration import GrpcConfiguration
+from momento.config.transport.transport_strategy import StaticGrpcConfiguration
 from momento.internal._utilities import _timedelta_to_ms
 
 DEFAULT_MAX_MESSAGE_SIZE = 5_243_000  # bytes
 
+ChannelArguments = Sequence[Tuple[str, Union[int, None]]]
```
anitarua, line 11: After refactoring into two methods, the linter complained:

> `src/momento/internal/_utilities/_grpc_channel_options.py:69:12: error: Expression has type "Any"`

So I made a type alias with `Any` in it that should be accepted by both async and sync versions

anitarua: This was after looking up that the grpc.aio return type was `(type alias) ChannelArgumentType: type[Sequence[Tuple[str, Any]]]`

malandis: Thanks for doing that. The Python gRPC library doesn't have type hints so we typically just add linter ignores to them in `pyproject.toml`. This is good too though!

malandis (approved):
🧼

PR #448, opened 2024-03-22
bumping the version in the example project

jersu11:
The version of the momento client-sdk-python in the poetry config was out of date, causing the example code to crash


malandis:
Thanks @jersu11 ! There is an automated task that bumps the version, but the CI failed due to a transient error -- see [here](https://github.com/momentohq/client-sdk-python/pull/447). I have bumped that so the version is now up to date. I appreciate your feedback 🙇

PR #449, opened 2024-04-03
fix: fix broken sdk link

virratanasangpunth:


cprice404:
Discussed with @virratanasangpunth IRL; this file is generated so this won't work. We need to fix the dev docs (add redirects, and the generator. PRs forthcoming).

PR #452, opened 2024-05-16
chore: add lambda zip example

malandis:
Adds a lambda example where we build the zip and upload to AWS via the
console.

Closes #819


examples/lambda/README.md
```diff
@@ -42,4 +42,24 @@ npm run cdk deploy
...
+
+1. Run `make dist` in the `zip` directory to package the lambda for upload as `dist.zip`.
+   :tipp: Check out the Makefile for important build steps to package for AWS Lambda.
+
+2. Create a new Lambda function by selecting "Author from scratch".
+3. Set the function name to `momento-lambda-demo`.
+4. Choose the runtime as `Python 3.8` (you can adjust this as desired).
+5. Select the architecture as `x86_64`.
+6. Click on "Create function" to create the Lambda function.
+7. In the "Code" tab, choose "Upload from the zip" as the code source.
```
cprice404, line 61: @eaddingtonwhite - do you have any input as to whether showing the steps to deploy via console will be sufficient for most users? In the one we just did for nodejs compression we showed how to deploy via `aws` cli and `sam`, but I don't have a good sense for whether users really need help with those parts once they have the zip part sorted.

cprice404 (commented):
Thanks for doing this! lgtm pending Ellery's input on whether we need to mention how to deploy with `aws` CLI and/or `sam`.

zip

examples/lambda/README.md
eaddingtonwhite: Yeah actually think console maybe better less moving parts or other steps people could mess up.

cprice404: okay gtk!

in that case, 🚢

cprice404 (approved):

PR #454, opened 2024-05-17
chore: use ubuntu-24.04 for ci where possible

malandis:
Updates the CI workflows to use `ubuntu-24.04` for python 3.9+ and `ubuntu-20.04` for python 3.7 and 3.8. Previously we could not run on `ubuntu-22.04` due to a difficult-to-diagnose timeout problem. This timeout problem manifested on Azure VMs with ubuntu-22.04 but not on other platforms.

This problem does not manifest on `ubuntu-24.04`, so we can use that where possible. Because python 3.7 and python 3.8 are not available on `ubuntu-24.04`, we still run those on `ubuntu-20.04`. At some point we can look in to installing via `pyenv` or manually on `ubuntu-24.04`.

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -44,7 +44,7 @@ jobs:
         uses: snok/install-poetry@v1
         with:
           version: 1.3.1
-          virtualenvs-in-project: false
+          virtualenvs-in-project: true
 
       - name: Setup Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/on-pull-request.yml
```diff
@@ -6,14 +6,27 @@ on:
...
+        # TODO: one of the examples dependencies does not support 3.11
+        os: [ubuntu-24.04]
+        python-version: ["3.9", "3.10", "3.11"]
         new-python-protobuf: ["true"]
         include:
+          # 3.7 and 3.8 are no longer available on ubuntu-24.04
+          # We run on 20.04 which was the last version where this worked.
+          # If support for 20.04 becomes an issue, we can install 3.7 and 3.8
+          # with pyenv or manually.
+          - python-version: "3.7"
```
bruuuuuuuce, line 20: interesting, I didn't know you could do this

bruuuuuuuce (approved):
:shipit:

cprice404 (approved):

PR #460, opened 2024-09-18
feat: add auth client and generate_disposable_token api

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1016

Will follow up with examples/docs updates.

To add the generate_disposable_token api, we needed to implement the AuthClient and permissions classes as well.

Note: Go sdk's permissions structure includes topic name prefix as well, but JS sdk does not. Should topic name prefix be added here?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ lint:
 
 .PHONY: do-gen-sync
 do-gen-sync:
+# cache client
 	@poetry run python src/momento/internal/codegen.py src/momento/internal/aio/_scs_control_client.py src/momento/internal/synchronous/_scs_control_client.py
 	@poetry run python src/momento/internal/codegen.py src/momento/internal/aio/_scs_data_client.py src/momento/internal/synchronous/_scs_data_client.py
 	@poetry run python src/momento/internal/codegen.py src/momento/cache_client_async.py src/momento/cache_client.py
@@ -37,6 +38,10 @@ do-gen-sync:
 	@poetry run python src/momento/internal/codegen.py tests/momento/cache_client/test_set_async.py tests/momento/cache_client/test_set.py
 	@poetry run python src/momento/internal/codegen.py tests/momento/cache_client/test_sorted_set_async.py tests/momento/cache_client/test_sorted_set.py
 	@poetry run python src/momento/internal/codegen.py tests/momento/cache_client/test_sorted_set_simple_async.py tests/momento/cache_client/test_sorted_set_simple.py
+# auth client
+	@poetry run python src/momento/internal/codegen.py src/momento/internal/aio/_scs_token_client.py src/momento/internal/synchronous/_scs_token_client.py
+	@poetry run python src/momento/internal/codegen.py src/momento/auth_client_async.py src/momento/auth_client.py
+	@poetry run python src/momento/internal/codegen.py tests/momento/auth_client/test_auth_client_async.py tests/momento/auth_client/test_auth_client.py
 
 .PHONY: gen-sync
 ## Generate synchronous code and tests from asynchronous code.
--- a/src/momento/auth/access_control/disposable_token_scope.py
+++ b/src/momento/auth/access_control/disposable_token_scope.py
@@ -1,0 +1,56 @@
+from dataclasses import dataclass
+from typing import List, Optional, Union
+
+from momento.auth.access_control.permission_scope import (
+    CachePermission,
+    CacheRole,
+    CacheSelector,
+    Permissions,
+)
+
+
+class AllCacheItems:
+    pass
+
+@dataclass
+class CacheItemKey:
+    key: str
+
+@dataclass
+class CacheItemKeyPrefix:
+    key_prefix: str
+
+@dataclass
+class CacheItemSelector:
+    cache_item: Union[CacheItemKey, CacheItemKeyPrefix, AllCacheItems, str]
+
+    def __init__(self, cache_item: Union[CacheItemKey, CacheItemKeyPrefix, AllCacheItems, str]):
+        self.cache_item = cache_item
+
+    def is_all_cache_items(self) -> bool:
+        return isinstance(self.cache_item, AllCacheItems)
+
+class DisposableTokenCachePermission(CachePermission):
+    cache_item_selector: CacheItemSelector
+
+    def __init__(self, item: CacheItemSelector, cache: CacheSelector, role: CacheRole):
+        super().__init__(cache, role)
+        self.item = item
+
+@dataclass
+class DisposableTokenCachePermissions:
+    disposable_token_permissions: List[DisposableTokenCachePermission]
+
+    def __init__(self, permissions: List[DisposableTokenCachePermission]):
+        self.disposable_token_permissions = permissions
+
+@dataclass
+class DisposableTokenScope:
+    disposable_token_scope: Union[Permissions, DisposableTokenCachePermissions]
+
+    def __init__(self, permission_scope: Union[Permissions, DisposableTokenCachePermissions]):
+        self.disposable_token_scope = permission_scope
+
+@dataclass
+class DisposableTokenProps:
+    token_id: Optional[str]
--- a/src/momento/auth/access_control/disposable_token_scope.py
+++ b/src/momento/auth/access_control/disposable_token_scope.py
@@ -12,13 +12,16 @@ from momento.auth.access_control.permission_scope import (
 class AllCacheItems:
     pass
 
+
 @dataclass
 class CacheItemKey:
     key: str
 
+
 @dataclass
 class CacheItemKeyPrefix:
     key_prefix: str
+
 
 @dataclass
 class CacheItemSelector:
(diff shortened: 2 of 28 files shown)


malandis:
Have you updated the `do-gen-sync` Makefile target to generate the synchronous code from the async code? Eg:
- `auth_client_async` should generate `auth_client`
- `scs_token_client` in the asynchronous internal directory should generate the corresponding synchronous one
- synchronous tests from async

anitarua:
> Have you updated the do-gen-sync Makefile target to generate the synchronous code from the async code? Eg:
> 
> - auth_client_async should generate auth_client
> - scs_token_client in the asynchronous internal directory should generate the corresponding synchronous one
> - synchronous tests from async

Oh heck, don't think I have yet. Thanks for pointing that out!

src/momento/auth/access_control/disposable_token_scope.py
```diff
@@ -0,0 +1,68 @@
...
+class CacheItemKeyPrefix:
+    key_prefix: str
+
+
+@dataclass
+class CacheItemSelector:
+    cache_item: Union[CacheItemKey, CacheItemKeyPrefix, AllCacheItems, str]
+
+    def __init__(self, cache_item: Union[CacheItemKey, CacheItemKeyPrefix, AllCacheItems, str]):
+        self.cache_item = cache_item
```
malandis, line 31: Tip: the dataclass decorator generates the init method based on the class variables you put above so this isn't necessary in any of the dataclasses (here and elsewhere). It generates other things as well such as `__eq__`, `__repr__`, `__hash__`, etc.

src/momento/auth/access_control/permission_scope.py
```diff
@@ -0,0 +1,110 @@
...
+class AllCaches:
+    pass
+
+
+class AllTopics:
+    pass
+
+
+class PredefinedScope:
+    pass
```
malandis, line 17: Can make these dataclasses as well to get the `__repr__` and `__str__` methods.

src/momento/responses/auth/generate_disposable_token.py
```diff
@@ -0,0 +1,87 @@
...
+        endpoint: str
+        """The endpoint the Momento client should use when making requests."""
+
+        expires_at: ExpiresAt
+        """The time at which the disposable token will expire."""
+
+        def __init__(self, auth_token: str, endpoint: str, expires_at: ExpiresAt):
+            self.auth_token = auth_token
+            self.endpoint = endpoint
+            self.expires_at = expires_at
```
malandis, line 63: I think you meant to make this a `dataclass`. Otherwise the variables you see here (`auth_token`, `endpoint`, `expires_at`) are _class variables_, not instance variables. They will be like static fields from other languages, which isn't what we want here. The `dataclass` decorator takes any class variable, turns it into an instance variable, and adds it to the generated `__init__`.

src/momento/utilities/expiration.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import Optional
+
+
+class Expiration(ABC):
+    """Represents an expiration time for a token."""
+
+    _does_expire: bool
+
+    def __init__(self, does_expire: bool) -> None:
+        self._does_expire = does_expire
```
malandis, line 15: Same situation in this file: `_does_expire` declared like this is a class variable, but you want an instance variable. You should omit it.

src/momento/utilities/expiration.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @abstractmethod
+    def does_expire(self) -> bool:
+        pass
```
malandis, line 19: The `Expiration` class can be a normal base class, as opposed to an `ABC`. This method doesn't need to be abstract since the derived classes implement in the same way. We can lift that logic up here.

Once you do that you can just make `Expiration` a dataclass.

src/momento/utilities/expiration.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class ExpiresIn(Expiration):
+    """Represents an expiration time for a token that expires in a certain amount of time."""
+
+    # Must be a float in order to use math.inf to indicate non-expiration
+    validFor: float
```
malandis, line 26: Same comment here re: class variables.

src/momento/utilities/expiration.py
```diff
@@ -0,0 +1,96 @@
...
+    def days(validForDays: int) -> ExpiresIn:
+        """Constructs an ExpiresIn with a specified validFor period in days."""
+        return ExpiresIn(validForDays * 86400)
+
+    @staticmethod
+    def epoch(expiresBy: int) -> ExpiresIn:
+        """Constructs an ExpiresIn with a specified expiresBy period in epoch format."""
+        current_epoch = int(datetime.now().timestamp())
+        seconds_until_epoch = expiresBy - current_epoch
+        return ExpiresIn(seconds_until_epoch)
```
malandis, line 70: style nit: use snake case for variables. Some camel case snuck in here.

src/momento/utilities/expiration.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class ExpiresAt(Expiration):
+    """Represents an expiration time for a token that expires at a certain UNIX epoch timestamp."""
+
+    # Must be a float in order to use math.inf to indicate non-expiration
+    expires_at: float
+
```
malandis, line 78: Same remark re: class variables

malandis (commented):
Few remarks re: class syntax

src/momento/internal/_utilities/_permissions.py
```diff
@@ -0,0 +1,225 @@
...
+    PredefinedScope,
+    TopicName,
+    TopicPermission,
+    TopicRole,
+)
+from tests.utils import str_to_bytes
+
+
+class SuperuserPermissions(PredefinedScope):
+    pass
```
anitarua, line 28: This class is private in the JS sdk and I tried to make it private / internal-only here but it's still importable when trying it out from the examples directory. 
I can also just omit this class for now since it's not used for disposable tokens

Makefile
```diff
@@ -37,6 +38,10 @@ do-gen-sync:
 	@poetry run python src/momento/internal/codegen.py tests/momento/cache_client/test_set_async.py tests/momento/cache_client/test_set.py
 	@poetry run python src/momento/internal/codegen.py tests/momento/cache_client/test_sorted_set_async.py tests/momento/cache_client/test_sorted_set.py
 	@poetry run python src/momento/internal/codegen.py tests/momento/cache_client/test_sorted_set_simple_async.py tests/momento/cache_client/test_sorted_set_simple.py
+# auth client
+	@poetry run python src/momento/internal/codegen.py src/momento/internal/aio/_scs_token_client.py src/momento/internal/synchronous/_scs_token_client.py
+	@poetry run python src/momento/internal/codegen.py src/momento/auth_client_async.py src/momento/auth_client.py
+	@poetry run python src/momento/internal/codegen.py tests/momento/auth_client/test_auth_client_async.py tests/momento/auth_client/test_auth_client.py
```
malandis, line 44: 👍

src/momento/auth_client_async.py
```diff
@@ -0,0 +1,88 @@
...
+        self,
+        scope: DisposableTokenScope,
+        expires_in: ExpiresIn,
+        disposable_token_props: Optional[DisposableTokenProps] = None,
+    ) -> GenerateDisposableTokenResponse:
+        """Generate a disposable auth token.
+
+        Returns:
+            GenerateDisposableTokenResponse
+        """
```
malandis, line 82: Let's document the parameters as well.

src/momento/auth/access_control/disposable_token_scopes.py
```diff
@@ -0,0 +1,206 @@
...
+    Permissions,
+    TopicName,
+    TopicPermission,
+    TopicRole,
+    TopicSelector,
+)
+
+
+class DisposableTokenScopes:
+    @staticmethod
```
malandis, line 31: The classes and methods here need docstrings

src/momento/auth/access_control/permission_scope.py
```diff
@@ -0,0 +1,101 @@
...
+Defines the data classes for specifying the permission scope of a generated token.
+"""
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Union
+
+
+@dataclass
+class AllCaches:
+    pass
```
malandis, line 12: These classes, methods, and fields need docstrings as well

src/momento/auth/access_control/permission_scopes.py
```diff
@@ -0,0 +1,103 @@
...
+    TopicPermission,
+    TopicRole,
+    TopicSelector,
+)
+
+
+class PermissionScopes:
+    @staticmethod
+    def cache_read_write(cache: Union[AllCaches, CacheName, str]) -> PermissionScope:
+        scope = Permissions(
```
malandis, line 26: Docstrings here too

malandis (commented):
Looks like we're almost there. Just need docstrings on a few of the dataclasses.

malandis (approved):

PR #462, opened 2024-09-19
chore: fall back manual version when dist-info absent

malandis:
We temporarily fall back to a manually provided version number when
the `dist-info` METADATA is absent. This situation occurs in some
frameworks like `serverless` which strip this for a slim build.

In the near future we will refactor this code so that CI sets the
version number in the file and we do not rely on metadata reflection.


cprice404 (approved):
do it

PR #464, opened 2024-09-23
fix: do not use test util in package

anitarua:
Ran into an error while writing new dev docs examples. Created a new `str_to_bytes` util function in the package proper rather than using the one from `tests.utils`. Bug appears to be fixed when running the examples against a locally built version of the package now.

```
File "/Users/anita/client-sdk-python/examples/.venv/lib/python3.10/site-packages/momento/internal/_utilities/_permissions.py", line 28, in <module>
    from tests.utils import str_to_bytes
ModuleNotFoundError: No module named 'tests'
```

Also moved the two validators to internal utils where the others are.

### DIFF
--- a/src/momento/cache_client.py
+++ b/src/momento/cache_client.py
@@ -8,10 +8,10 @@ from momento import logs
 from momento.auth import CredentialProvider
 from momento.config import Configuration
 from momento.errors import UnknownException
+from momento.internal._utilities._data_validation import validate_eager_connection_timeout
 from momento.requests import CollectionTtl, SortOrder
 from momento.utilities.shared_sync_asyncio import (
     DEFAULT_EAGER_CONNECTION_TIMEOUT_SECONDS,
-    validate_eager_connection_timeout,
 )
 
 try:
--- a/src/momento/cache_client.py
+++ b/src/momento/cache_client.py
@@ -8,7 +8,7 @@ from momento import logs
 from momento.auth import CredentialProvider
 from momento.config import Configuration
 from momento.errors import UnknownException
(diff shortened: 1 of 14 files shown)


rishtigupta (approved):
:shipit:

malandis:
While we're here, the validators are best placed alongside the ones [here](https://github.com/momentohq/client-sdk-python/tree/main/src/momento/internal/_utilities)

malandis:
Lastly could you roll up the names from the utilities to the utilities`__init__`? Should make the imports easier throughout the project.

src/momento/cache_client.py
```diff
@@ -8,10 +8,10 @@
 from momento.auth import CredentialProvider
 from momento.config import Configuration
 from momento.errors import UnknownException
+from momento.internal._utilities._data_validation import _validate_eager_connection_timeout
```
malandis, line 11:
```suggestion
from momento.internal._utilities import _validate_eager_connection_timeout
```

malandis (commented):
Also need the init file for `utilities` that rolls up names from the various modules therein.

src/momento/utilities/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .expiration import Expiration, ExpiresAt, ExpiresIn
+from .shared_sync_asyncio import DEFAULT_EAGER_CONNECTION_TIMEOUT_SECONDS, str_to_bytes
```
malandis, line 2: You'll also need an [`__all__` definition](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) with the names here

malandis (approved):

PR #466, opened 2024-09-24
chore: add doc examples for auth client

anitarua:


examples/py310/doc-examples-python-apis.py
```diff
@@ -19,6 +22,8 @@
     TopicSubscribe,
     TopicSubscriptionItem,
 )
+from momento.responses.auth.generate_disposable_token import GenerateDisposableToken
+from momento.utilities.expiration import ExpiresIn
```
malandis, line 26:
```suggestion
from momento.utilities import ExpiresIn
```

examples/py310/doc-examples-python-apis.py
```diff
@@ -19,6 +22,8 @@
     TopicSubscribe,
     TopicSubscriptionItem,
 )
+from momento.responses.auth.generate_disposable_token import GenerateDisposableToken
```
malandis, line 25: Not a blocker but did we roll up the various types under the auth modules to the auth init? That way we wouldn't have to reach into `generate_disposable_token`?

malandis (commented):
import nit

examples/py310/doc-examples-python-apis.py
anitarua: ah heck, I had rolled up `GenerateDisposableTokenResponse` but not `GenerateDisposableToken`. I can make a small PR for that change

anitarua: That'll be fixed here: https://github.com/momentohq/client-sdk-python/pull/468

malandis (approved):

PR #487, opened 2024-11-08
chore: specify path to release-please manifest

anitarua:
Follow up to #486 -- release-please still did not produce a new PR to cut a new release.

~~I think switching the release-please github actions version to `v3` to match what the other SDKs use will allow the `changelog-types` field to be picked up and parse our commits correctly.~~

~~Update: using `v4` shouldn't be a problem. I think we're just missing the argument to specify the path to the manifest, as shown in their [example](https://github.com/googleapis/release-please-action?tab=readme-ov-file#advanced-release-configuration).~~

Found an [issue](https://github.com/googleapis/release-please-action/issues/941) in the release-please action that seems to show the same problem (action ends with "No user facing commits found since ... - skipping"), and the solution was to move all the configs into the manifest. Our file still had `release-type` as part of the github action yaml block.

### DIFF
--- a/.github/workflows/on-push-to-main-branch.yml
+++ b/.github/workflows/on-push-to-main-branch.yml
@@ -35,14 +35,13 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: googleapis/release-please-action@v3
+      - uses: googleapis/release-please-action@v4
         name: Release Please
         id: release
         with:
           token: ${{ secrets.MOMENTO_MACHINE_USER_GITHUB_TOKEN }}
           release-type: python
-          default-branch: main
-          changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"chore","section":"Miscellaneous","hidden":false}]'
+          manifest-file: .release-please-manifest.json
 
   publish:
     needs: [release-please]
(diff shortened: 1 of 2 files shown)


malandis (commented):
I think there's a misunderstanding. In v4 the changelog types have to be defined in `.release-please-manifest.json` (see [ours](https://github.com/momentohq/client-sdk-python/blob/main/.release-please-manifest.json).

Let's dig in to this on Monday

.github/workflows/on-push-to-main-branch.yml
```diff
@@ -41,8 +41,7 @@ jobs:
         with:
           token: ${{ secrets.MOMENTO_MACHINE_USER_GITHUB_TOKEN }}
           release-type: python
-          default-branch: main
-          changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"chore","section":"Miscellaneous","hidden":false}]'
+          manifest-file: .release-please-manifest.json
```
malandis, line 44: I don't think this is the root cause: `manifest-file` defaults to this already ("manifest-file | Path to the release-please versions manifest. Defaults to .release-please-manifest.json"). Also, I added `momento/__init__.py` as an extra file to modify and it did that

malandis (commented):
Looks like we'll have to dig in further. We can sync about it.

malandis (approved):

PR #495, opened 2024-12-05
chore: update list of retryable gRPC functions

nand4011:


src/momento/retry/default_eligibility_strategy.py
```diff
@@ -28,32 +28,57 @@
 ]
 
 RETRYABLE_REQUEST_TYPES: list[str] = [
```
nand4011, line 30: Compare to the [Go SDK version](https://github.com/momentohq/client-sdk-go/blob/9ebbbe63aefbfdfee3c0fcbe574973980838b54b/internal/retry/eligibility_strategy.go#L16).

malandis (commented):
Both of the types there should be Python `set`s. Otherwise looks good

malandis (approved):

PR #496, opened 2025-03-04
feat: add topic grpc config and transport strategy and make sure publish deadline is set

anitarua:
Closes [Issue #1133](https://github.com/momentohq/dev-eco-issue-tracker/issues/1133)

Publish deadline was not getting set. This PR fixes that and adds a TopicGrpcConfiguration and TopicTransportStrategy to make the publish timeout configurable.

### DIFF
--- a/src/momento/config/topic_configurations.py
+++ b/src/momento/config/topic_configurations.py
@@ -26,6 +26,6 @@ class TopicConfigurations:
             This configuration is guaranteed not to change in future releases of the Momento Python SDK.
             """
             return TopicConfigurations.Default(
-                StaticTopicTransportStrategy(StaticTopicGrpcConfiguration(deadline=timedelta(seconds=10))),
+                StaticTopicTransportStrategy(StaticTopicGrpcConfiguration(deadline=timedelta(seconds=5))),
                 max_subscriptions=0,
             )
--- a/src/momento/config/topic_configurations.py
+++ b/src/momento/config/topic_configurations.py
@@ -1,5 +1,9 @@
 from __future__ import annotations
 
+from datetime import timedelta
+
+from momento.config.transport.topic_transport_strategy import StaticTopicGrpcConfiguration, StaticTopicTransportStrategy
+
 from .topic_configuration import TopicConfiguration
 
 
@@ -21,4 +25,7 @@ class TopicConfigurations:
 
             This configuration is guaranteed not to change in future releases of the Momento Python SDK.
             """
-            return TopicConfigurations.Default(max_subscriptions=0)
+            return TopicConfigurations.Default(
+                StaticTopicTransportStrategy(StaticTopicGrpcConfiguration(deadline=timedelta(seconds=5))),
+                max_subscriptions=0,
+            )
--- a/src/momento/config/topic_configurations.py
+++ b/src/momento/config/topic_configurations.py
@@ -26,6 +26,6 @@ class TopicConfigurations:
             This configuration is guaranteed not to change in future releases of the Momento Python SDK.
             """
             return TopicConfigurations.Default(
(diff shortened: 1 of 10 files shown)


anitarua:
Hm, should we remove ubuntu-20.04 from github actions workflow? 

> This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner will be removed on 2025-04-01. For more details, see https://github.com/actions/runner-images/issues/11101

src/momento/config/topic_configurations.py
```diff
@@ -21,4 +25,7 @@ def v1() -> TopicConfigurations.Default:
 
             This configuration is guaranteed not to change in future releases of the Momento Python SDK.
             """
-            return TopicConfigurations.Default(max_subscriptions=0)
+            return TopicConfigurations.Default(
+                StaticTopicTransportStrategy(StaticTopicGrpcConfiguration(deadline=timedelta(seconds=10))),
```
malandis, line 29: In the previous commit, the deadline was 1100ms. Are we going we 10s across the clients?

anitarua: Previously, the deadline wasn't being set on the request at all, so the 1100ms wasn't being applied. 

I tried a lower default deadline (5s) while developing and observed some failing tests. Double checked and it seems deadline of 6s or above would pass those tests though.

<details>
<summary>3 tests fail with 1100ms default deadline</summary>

```
FAILED tests/momento/topic_client/test_topics.py::describe_publish::a_cache_name_validator::with_non_existent_cache_name_it_throws_not_found - AssertionError
FAILED tests/momento/topic_client/test_topics_async.py::describe_publish::a_cache_name_validator::with_non_existent_cache_name_it_throws_not_found - AssertionError
FAILED tests/momento/topic_client/test_topics_async.py::describe_publish::publish_happy_path - assert False

Results (20.14s):
      29 passed
       3 failed
         - tests/momento/topic_client/shared_behaviors.py:18 a_cache_name_validator::with_non_existent_cache_name_it_throws_not_found
         - tests/momento/topic_client/shared_behaviors_async.py:20 a_cache_name_validator::with_non_existent_cache_name_it_throws_not_found
         - tests/momento/topic_client/test_topics_async.py:32 publish_happy_path
```

</details>

<details>
<summary>2 tests fail with 5s default deadline</summary>

```
FAILED tests/momento/topic_client/test_topics.py::describe_publish::a_cache_name_validator::with_non_existent_cache_name_it_throws_not_found - AssertionError
FAILED tests/momento/topic_client/test_topics_async.py::describe_publish::a_cache_name_validator::with_non_existent_cache_name_it_throws_not_found - AssertionError
```

</details>

anitarua: the default timeout in nodejs appears to be 5s ([reference](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/client-sdk-nodejs/src/internal/pubsub-client.ts#L36))

malandis: The failures on 5s are suspicious since they have bogus inputs. Those same two tests also failed when using a 1100ms deadline.

malandis: Regardless, we should have a consistent timeout across the SDKs. If we need the tests to have a higher timeout to be generous in CI, then we should configure the test fixture to have a higher time out. We should also investigate why these two tests take longer and if that is a service issue.

anitarua: sounds good, I'll make the default timeout 5s to match the other sdks, and just bump up the timeout on the tests.
I can also take a quick look into why those two tests take longer (maybe due to the`@behaves_like` and `@fixture` stuff?)

anitarua: hmm, those tests are passing locally for me today, even when using 1100ms 🙃

malandis (approved):

PR #498, opened 2025-03-05
ci: remove testing python 3.7 and 3.8, add testing python 3.12

anitarua:
We had received the warning:

> This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner will be removed on 2025-04-01. For more details, see https://github.com/actions/runner-images/issues/11101

Michael previously found no easy way to install python 3.7 and 3.8 on Ubuntu 24.04, and once we confirm no customers are still using those versions, we plan to remove 3.7 and 3.8 from the test matrix. 

This PR will also update the test matrix to include python 3.12 and 3.13

Edit: it seems adding 3.13 won't work. The package [grpcio didn't support python 3.13 until v1.66.2](https://github.com/grpc/grpc/issues/34922#issuecomment-2380652812) and attempting to upgrade grpcio to that version produces this message:

```
The current project's Python requirement (>=3.7,<4.0) is not compatible with some of the required packages Python requirement:
  - grpcio requires Python >=3.8, so it will not be satisfied for Python >=3.7,<3.8

Because momento depends on grpcio (1.66.2) which requires Python >=3.8, version solving failed.
```

And in CI, python 3.13 cannot install seem to install the version of grpcio currently used in the sdk.


malandis:
Let's remove 3.13 from the test matrix for now. Can follow up on resolving the version conflict with grpcio later. 👍

malandis:
I added support for python 3.13 in [5ee1847](https://github.com/momentohq/client-sdk-python/pull/498/commits/5ee18479d68208748be7dd977aa60cba4bdd6bf3). Once we merge this we can bump the examples version.

pgautier404 (approved):
LGTM!

PR #503, opened 2025-04-09
chore: add tests for the fixed count retry strategy

nand4011:
Add a middleware that lets tests control Momento Local with metadata and can collect metrics about retries.

Add sync and async tests for the fixed count retry strategy.

Update the aio retry interceptor to stop if from ignoring the result of a future in some cases.

Add a make target for local tests.

Add a github action to run the local tests.

### DIFF
--- a/.github/workflows/momento-local-tests.yml
+++ b/.github/workflows/momento-local-tests.yml
@@ -1,0 +1,41 @@
+name: Momento Local tests
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  local-tests:
+    strategy:
+      matrix:
+        os: [ubuntu-24.04]
+        python-version: ["3.13"]
+    runs-on: ${{ matrix.os }}
+
+    env:
+      TEST_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
+      TEST_CACHE_NAME: python-integration-test-${{ matrix.python-version }}-${{ matrix.new-python-protobuf }}-${{ github.sha }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install and configure Poetry
+        uses: snok/install-poetry@v1
+        with:
+          version: 1.3.1
+          virtualenvs-in-project: true
+
+      - name: Install dependencies
+        run: poetry install
+
+      - name: Start Momento Local
+        run: |
+          docker run --cap-add=NET_ADMIN --rm -d -p 8080:8080 -p 9090:9090 gomomento/momento-local --enable-test-admin
+
+      - name: Run tests
+        run: poetry run pytest -p no:sugar -q -m local
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,12 @@ gen-sync: do-gen-sync format lint
 .PHONY: test
 ## Run unit and integration tests with pytest
 test:
-	@poetry run pytest
+	@poetry run pytest -m "not local"
+
+.PHONY: test-local
+## Run the integration tests that require Momento Local
+test-local:
+	@poetry run pytest -m local
 
 .PHONY: precommit
 ## Run format, lint, and test as a step before committing.
--- a/tests/momento/local/test_fixed_count_retry_strategy.py
+++ b/tests/momento/local/test_fixed_count_retry_strategy.py
@@ -1,0 +1,71 @@
+import pytest
+from momento.errors import MomentoErrorCode
+from momento.responses import CacheGet, CacheIncrement
+
+from tests.conftest import client_local
+from tests.momento.local.momento_local_async_middleware import MomentoLocalMiddlewareArgs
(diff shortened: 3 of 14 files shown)


.github/workflows/momento-local-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  local-tests:
+    strategy:
+      matrix:
+        os: [ubuntu-24.04]
+        python-version: ["3.13"]
+    runs-on: ${{ matrix.os }}
+
+    env:
+      TEST_API_KEY: ${{ secrets.ALPHA_TEST_AUTH_TOKEN }}
```
anitarua, line 16: no key needed

.github/workflows/momento-local-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Install and configure Poetry
+        uses: snok/install-poetry@v1
+        with:
+          version: 1.3.1
+          virtualenvs-in-project: true
+
+      - name: Install dependencies
+        run: poetry install
+
+      - name: Run tests
```
anitarua, line 36: need to start docker before running tests, no?

`docker run --cap-add=NET_ADMIN --rm -d -p 8080:8080 -p 9090:9090 gomomento/momento-local --enable-test-admin` will cover both retry strategy and topics retry tests

Makefile
```diff
@@ -50,7 +50,12 @@ gen-sync: do-gen-sync format lint
 .PHONY: test
 ## Run unit and integration tests with pytest
 test:
-	@poetry run pytest
+	@poetry run pytest -m "not local"
+
+.PHONY: test-local
+## Run the integration tests that require Momento Local
+test-local:
```
anitarua, line 57: nit: naming inconsistency with other sdks

```suggestion
test-retry:
```

src/momento/config/configuration.py
```diff
@@ -106,10 +131,60 @@ def with_root_certificates_pem(self, root_certificates_pem_path: Path) -> Config
...
-            ConfigurationBase: the new Configuration.
+            Configuration: the new Configuration.
         """
         grpc_configuration = self._transport_strategy.get_grpc_configuration().with_root_certificates_pem(
             root_certificates_pem_path
         )
         transport_strategy = self._transport_strategy.with_grpc_configuration(grpc_configuration)
         return self.with_transport_strategy(transport_strategy)
+
+    def with_middlewares(self, middlewares: List[Middleware]) -> Configuration:
```
pgautier404, line 142: I believe the intent of this is to *replace* the existing middleware list instead of extending it?

.github/workflows/momento-local-tests.yml
nand4011: Unfortunately a valid key is needed for the common logic that the tests use. I could potentially change that, but I'd prefer to do it in a follow-on pr.

Makefile
nand4011: I did that because we named them retry tests because retries were the first thing we did with momento local, and I figured it would be more general in the future. I made a bunch of retry -> momentoLocal name changes in the Java test logic, for example.
I can make it test-retry if we value the consistency though.

anitarua (dismissed):
lgtm!

tests/momento/local/test_fixed_count_retry_strategy.py
```diff
@@ -0,0 +1,71 @@
...
+from momento.responses import CacheGet, CacheIncrement
+
+from tests.conftest import client_local
+from tests.momento.local.momento_local_async_middleware import MomentoLocalMiddlewareArgs
+from tests.momento.local.momento_local_metrics_collector import MomentoLocalMetricsCollector
+from tests.momento.local.momento_rpc_method import MomentoRpcMethod
+from tests.utils import uuid_str
+
+
+class TestFixedCountRetryStrategyAsync:
```
malandis, line 12: Why are these tests in a class vs as module-level functions?

tests/momento/local/test_fixed_count_retry_strategy_async.py
```diff
@@ -0,0 +1,74 @@
...
+from momento.responses import CacheGet, CacheIncrement
+
+from tests.conftest import client_async_local
+from tests.momento.local.momento_local_async_middleware import MomentoLocalMiddlewareArgs
+from tests.momento.local.momento_local_metrics_collector import MomentoLocalMetricsCollector
+from tests.momento.local.momento_rpc_method import MomentoRpcMethod
+from tests.utils import uuid_str
+
+
+class TestFixedCountRetryStrategyAsync:
```
malandis, line 12: Same question re: class. In pytest the tests are functions by convention.

Makefile
```diff
@@ -50,7 +50,12 @@ gen-sync: do-gen-sync format lint
...
+	@poetry run pytest -m local
```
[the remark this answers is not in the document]
nand4011: Part of my dilemma when choosing a name for this was that we'll have tests like these that always run against momento local, and the standard integration tests that could run with it. I wasn't sure whether we should separate those. We can solve that problem in the future, I suppose.

tests/momento/local/test_fixed_count_retry_strategy.py
nand4011: Because I'm not very pythonic :-). I'll update them.

malandis (approved):

PR #504, opened 2025-04-11
chore: fixed timeout retry strategy

pgautier404:


src/momento/internal/aio/_retry_interceptor.py
```diff
@@ -34,20 +35,43 @@ async def intercept_unary_unary(
         client_call_details: grpc.aio._interceptor.ClientCallDetails,
         request: grpc.aio._typing.RequestType,
     ) -> grpc.aio._call.UnaryUnaryCall | grpc.aio._typing.ResponseType:
+        call = None
         attempt_number = 1
+        # the overall deadline is calculated from the timeout set on the client call details
+        overall_deadline = datetime.now() + timedelta(seconds=client_call_details.timeout or 0.0)
```
malandis, line 41: When will `client_call_details` be `None`? If never for our clients we should add a comment that this is just for typing.

malandis (commented):
lgtm just one question re: defaults

malandis (approved):

PR #505, opened 2025-04-15
chore: remove configuration bases and make middleware more pythonic

nand4011:
Remove the configuration base classes. These aren't used by any of the code that takes a configuration, so users can't implement new configurations with them and pass them into clients. I'm removing them instead of making the clients use them because config base classes make it difficult to add new configuration options breaking potential user implementations.

Make the middleware classes into data classes. The existing standard classes with getters isn't pythonic.

### DIFF
--- a/src/momento/config/auth_configuration.py
+++ b/src/momento/config/auth_configuration.py
@@ -1,36 +1,13 @@
 from __future__ import annotations
 
-from abc import ABC, abstractmethod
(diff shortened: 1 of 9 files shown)


pgautier404 (dismissed):
This looks good to me as long as @malandis signs off.

malandis (approved):

PR #507, opened 2025-05-01
chore: send only-once headers per client instead of just once

nand4011:
Send the only-once headers once per instance of the add header interceptor, instead of once over the life of the program. The old behavior would make a cache client stop a topics client from sending the agent header, and vice versa.

### DIFF
--- a/src/momento/internal/aio/_add_header_client_interceptor.py
+++ b/src/momento/internal/aio/_add_header_client_interceptor.py
@@ -16,9 +16,8 @@ class Header:
 
 
 class AddHeaderStreamingClientInterceptor(grpc.aio.UnaryStreamClientInterceptor):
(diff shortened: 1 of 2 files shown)


src/momento/internal/aio/_add_header_client_interceptor.py
```diff
@@ -16,9 +16,8 @@ def __init__(self, name: str, value: str):
 
 
 class AddHeaderStreamingClientInterceptor(grpc.aio.UnaryStreamClientInterceptor):
-    are_only_once_headers_sent = False
-
```
malandis, line 20: Good find. These were class variables like this. If this were a data class, they would be instance variables.

malandis (approved):

PR #510, opened 2025-06-11
fix: subscriptions bookkeeping in pubsub clients

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1157

Pubsub clients should do some bookkeeping of number of active subscriptions to be able to gracefully reject new subscribe requests rather than letting them silently queue up on the client.

Updated the client to use a pool of 4 unary grpc managers by default to match the other SDKs. 
The stream managers use the existing `with_max_subscriptions` configuration to statically set the number of grpc managers in the stream pool.

Created a sync and async `subscription_initialization` test suite like in other SDKs. Unfortunately, it appears these tests cannot be run against momento-local. The credential provider must use `insecure_channel()` for http connections to momento-local, but apparently under the hood that means only one underlying TCP connection is created, but the tests are written expecting to exercise multiple grpc channels. I've excluded these tests from CI because they cannot be run against momento-local or the live service since they require a subscriptions limit greater than the default.


src/momento/responses/pubsub/subscribe.py
```diff
@@ -138,14 +157,33 @@ async def __anext__(self) -> TopicSubscriptionItemResponse:
...
+            self._decrement_stream_count = decrement_stream_count_method
+            self._decremented = False
 
         def __iter__(self) -> TopicSubscribe.Subscription:
             return self
 
+        def __del__(self) -> None:
+            if not self._decremented:
+                self._decrement_stream_count()
+                self._decremented = True
```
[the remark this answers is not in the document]
anitarua: I implemented `__del__` in order to be able to decrement the number of active streams when a subscription closes. Is there a better method to implement this? Or is there a `super()` equivalent I should call here?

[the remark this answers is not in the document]
malandis: The pattern for this is to use a [context manager](https://docs.python.org/3/library/contextlib.html). I don't think we can do this now though since it would break the API and expectations existing users have.

src/momento/responses/pubsub/subscribe.py
```diff
@@ -138,14 +157,33 @@ async def __anext__(self) -> TopicSubscriptionItemResponse:
...
+
+        def unsubscribe(self) -> None:
+            if not self._decremented:
+                self._decrement_stream_count()
+                self._decremented = True
+            self._client_stream.cancel()  # type: ignore[misc]
```
malandis, line 185: Does this need to be public?

malandis, line 185: Does this need to be public?

anitarua: Yeah, it's called in the tests / can be called by end users to immediately free up a stream

malandis (approved):

